PR #4, opened 2022-10-27
Support basic authentication via encoded Authorization header, as cur…

mseaton:
…rently done in FHIR and REST modules.


mogoodrich (approved):
LGTM, thanks for the quick response @mseaton !

PR #6, opened 2022-11-18
UHM-6854 - FHIR Basic Authentication Fails (in some cases?) when Auth…

mseaton:
…entication module is installed


mogoodrich (approved):
Oh, was I just doing this myself locally to test, will merge this in and test.

PR #7, opened 2022-11-18
Set login on thread in WebAuthenticationScheme if missing, for exampl…

mseaton:
…e if Context.authenticate is called directly.


mogoodrich (approved):
LGTM, thanks @mseaton , going to merge now so I can test...

PR #10, opened 2022-12-05
Add a filter to clear a configured list of cookies if a web session is invalidated

ibacher:
This adds a filter to this module that will clear any configured cookies if the user's session is expired during the request. The idea is that if the user's request goes to a logoff endpoint (regardless of what UI its in), we can clear any session-related cookies.

This is intended to resolve issue like in [this Talk post](https://talk.openmrs.org/t/client-cookies-are-not-getting-removed-after-server-session-invalidated/38047).

### DIFF
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -113,11 +113,6 @@ public class AuthenticationFilter implements Filter {
 			userLogin.setLastActivityDate(new Date());
 
 			if (!session.isUserAuthenticated()) {
-
-				if (!AuthenticationConfig.isConfigurationCacheEnabled()) {
-					AuthenticationConfig.reloadConfigFromRuntimeProperties(WebConstants.WEBAPP_NAME);
-				}
-
 				AuthenticationScheme authenticationScheme = getAuthenticationScheme();
 
 				if (authenticationScheme instanceof WebAuthenticationScheme) {
@@ -134,7 +129,9 @@ public class AuthenticationFilter implements Filter {
 						if (credentials != null) {
 							try {
 								session.authenticate(webScheme, credentials);
-								session.regenerateHttpSession();  // Guard against session fixation attacks
+								if (!session.getHttpSession().isNew()) {  // Only regenerate the session for a client-provided session
+									session.regenerateHttpSession();  // Guard against session fixation attacks
+								}
 								session.refreshDefaultLocale(); // Refresh context locale after authentication
 								String successUrl = determineSuccessRedirectUrl(request);
 								response.sendRedirect(successUrl);
@@ -150,9 +147,7 @@ public class AuthenticationFilter implements Filter {
 				}
 			}
 
-			if (!response.isCommitted()) {
-				chain.doFilter(servletRequest, servletResponse);
-			}
+			chain.doFilter(servletRequest, servletResponse);
 		}
 		finally {
 			UserLoginTracker.removeLoginFromThread();
--- a/omod/src/main/java/org/openmrs/module/authentication/web/CookieClearingFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/CookieClearingFilter.java
@@ -1,0 +1,111 @@
+package org.openmrs.module.authentication.web;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import java.io.IOException;
+import java.util.Arrays;
+
+import org.apache.commons.lang3.StringUtils;
+import org.openmrs.module.authentication.AuthenticationConfig;
+import org.openmrs.web.WebConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This servlet filter exists to remove session cookies when a user logs out.
+ * <p/>
+ * This filter is configurable at runtime using the following properties:
+ * <ul>
+ *     <li><tt>authentication.cookies.clearOnLogout = true | false </tt> determines whether to clear cookies on logout. If
+ *     not set to <tt>true</tt>, this filter takes no actions.</li>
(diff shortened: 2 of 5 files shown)


omod/src/main/java/org/openmrs/module/authentication/web/CookieClearingFilter.java
```diff
@@ -0,0 +1,111 @@
...
+				AuthenticationConfig.reloadConfigFromRuntimeProperties(WebConstants.WEBAPP_NAME);
+			}
+			
+			// if an earlier filter has already written a response, we cannot do anything
+			if (response.isCommitted()) {
+				filterChain.doFilter(servletRequest, servletResponse);
+				return;
+			}
+			
+			boolean isEnabled = AuthenticationConfig.getBoolean("authentication.cookies.clearOnLogout", false);
```
mseaton, line 61: It seems unnecessary to have both a boolean "clearOnLogout" and a delimited "toClear" property.  Can we just infer that this is enabled if the cookies to clear is not blank, and disabled if it is blank?

omod/src/main/java/org/openmrs/module/authentication/web/CookieClearingFilter.java
```diff
@@ -0,0 +1,111 @@
...
+ * This filter is configurable at runtime using the following properties:
+ * <ul>
+ *     <li><tt>authentication.cookies.clearOnLogout = true | false </tt> determines whether to clear cookies on logout. If
+ *     not set to <tt>true</tt>, this filter takes no actions.</li>
+ *     <li><tt>authentication.cookies.toClear = comma separated list of cookies to clear</tt>
+ *     determines the cookies we will try to clear. If unset, no action will be taken. If set, the named cookies will only be
+ *     cleared if they are present on the incoming request.</li>
+ * </ul>
+ */
+public class CookieClearingFilter implements Filter {
```
mseaton, line 34: Wouldn't it be better to just add this to the core web api, either as an additional filter in the filter chain or as something done within the existing OpenmrsFilter, rather than having to build this into this (and any other similar) module?

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -113,11 +113,6 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 			userLogin.setLastActivityDate(new Date());
 
 			if (!session.isUserAuthenticated()) {
-
-				if (!AuthenticationConfig.isConfigurationCacheEnabled()) {
```
mseaton, line 117: I assume you are removing this because you added it to the CookieClearingFilter and it runs first?  That seems OK.  It's really only a developer / implementer tool to use in a dev/test environment.

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -134,7 +129,9 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 						if (credentials != null) {
 							try {
 								session.authenticate(webScheme, credentials);
-								session.regenerateHttpSession();  // Guard against session fixation attacks
+								if (!session.getHttpSession().isNew()) {  // Only regenerate the session for a client-provided session
+									session.regenerateHttpSession();  // Guard against session fixation attacks
+								}
```
mseaton, line 134: I'm not 100% sure I understand this or why it is necessary, can you explain a bit more?  The code will only reach this point if the user starts out unauthenticated and there is a successful authentication of credentials during the course of processing this request from credentials in the request.  Is there a situation where that might happen and we _don't_ want to regenerate the session?

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -150,9 +147,7 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 				}
 			}
 
-			if (!response.isCommitted()) {
-				chain.doFilter(servletRequest, servletResponse);
-			}
```
mseaton, line 155: Why are you removing this?  Is this incorrect?  The code above can result in a redirect issued on the response, in which case I feel like it should immediately return and not continue processing through the filter chain.  I've certainly run into errors without this check in place before.

omod/src/main/java/org/openmrs/module/authentication/web/CookieClearingFilter.java
```diff
@@ -0,0 +1,111 @@
...
+			// if an earlier filter has already written a response, we cannot do anything
+			if (response.isCommitted()) {
+				filterChain.doFilter(servletRequest, servletResponse);
+				return;
+			}
+			
+			boolean isEnabled = AuthenticationConfig.getBoolean("authentication.cookies.clearOnLogout", false);
+			String[] cookiesToClear = new String[0];
+			if (isEnabled) {
+				String cookiesToClearSetting = AuthenticationConfig.getProperty("authentication.cookies.toClear", "");
```
mseaton, line 64: It would seem like if our primary motivation is to ensure the JSESSIONID is cleared, that we just default to this behavior and then let implementations override that, rather than have to force this to be configured by everyone

mogoodrich:
I haven't reviewed (@mseaton lmk me know if you need me to take a closer look), but I did notice a case recently when using Postman when I had to manually clear out a JSESSION cookie that had a expired, which I didn't think I had to do in the past, so building this into the auth module would be good to do.

omod/src/main/java/org/openmrs/module/authentication/web/CookieClearingFilter.java
ibacher: Yeah, that makes sense. Originally, there was some other functionality here, but that's no longer the case.

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
ibacher: So the point of regenerating the session is to ensure that the authenticated session is not using a client-supplied session id (to prevent sticky sessions). `isNew()` essentially just allows us to check whether the current session has been communicated to the client... In the case of a new session, i.e., if [this call](https://github.com/openmrs/openmrs-module-authentication/blob/2fcb3d809ffb16070927d5322723d92e39036408/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationSession.java#L88) creates a new session, invalidating and creating the new session is unnecessary work. I'm happy to take this out, though.

omod/src/main/java/org/openmrs/module/authentication/web/CookieClearingFilter.java
ibacher: My intention was to make this essentially something that you would opt into rather than something that would just be on by default, but that's not essential.

PR #11, opened 2022-12-08
Support reset password workflow

mseaton:


omod/src/test/java/org/openmrs/module/authentication/web/AuthenticationFilterTest.java
```diff
@@ -178,13 +178,9 @@ public void shouldRedirectToRequestedPageIfAuthenticationFails() throws Exceptio
 	}
 
 	@Test
-	public void shouldWhiteListIfAnyPatternsMatchAndGetRequest() {
+	public void shouldWhiteListIfAnyPatternsMatchRequest() {
```
mogoodrich, line 181: shouldn't we be testing here that POSTs are whitelisted?

mogoodrich (approved):
LGTM, one comment but feel free to ignore, I'm probably missing something... :)

PR #13, opened 2024-06-26
AUT-1: Support for forcing users to change their password on login

gitcliff:
PR is adding the ability to set require password change on first login via the openmrs runtime properties

### DIFF
--- a/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
+++ b/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
@@ -75,7 +75,7 @@ public class AuthenticationConfig implements Serializable {
     /**
      * URLs that might need to be made accessible without go through the password change authentication.
      */
-    public static final String PASS_WORD_CHANGE_WHITE_LIST = "authentication.passWordChangeWhiteList";
+    public static final String PASSWORD_CHANGE_WHITE_LIST = "authentication.passwordChangeWhiteList";
 
     /**
      * All AuthenticationScheme instances must be configured with, at minimum, a property that maps a particular
@@ -244,13 +244,13 @@ public class AuthenticationConfig implements Serializable {
      * @return the List of url patterns to allow without force password authentication redirection
      */
     public static List<String> getPasswordChangeWhiteList() {
-        List<String> whiteList = getStringList(PASS_WORD_CHANGE_WHITE_LIST);
+        List<String> whiteList = getStringList(PASSWORD_CHANGE_WHITE_LIST);
         whiteList.add(getChangePasswordUrl());  // Add the change password URL to the whitelist
         return whiteList;
     }
 
     public static String getChangePasswordUrl() {
-        return getConfig().getProperty(PASSWORD_CHANGE_URL, "/changePassword");
+        return getConfig().getProperty(PASSWORD_CHANGE_URL);
     }
 
 
--- a/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
+++ b/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
@@ -91,7 +91,6 @@ public class AuthenticationConfig implements Serializable {
 
     private static Properties config;
 
-
     private static final List<ClassLoader> classLoaders = new ArrayList<>();
 
     /**
--- a/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
+++ b/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
@@ -33,6 +33,8 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String PREFIX = "authentication";
 
+    private static Properties config;
+
     /**
      * All configured authentication schemes are identified by a unique {schemeId} in the configuration
      */
@@ -71,6 +73,11 @@ public class AuthenticationConfig implements Serializable {
     public static final String WHITE_LIST = "authentication.whiteList";
 
     /**
+     * URLs that might need to be made accessible without go through the password change authentication.
+     */
+    public static final String PASS_WORD_CHANGE_WHITE_LIST = "authentication.passWordChangeWhiteList";
+
+    /**
      * All AuthenticationScheme instances must be configured with, at minimum, a property that maps a particular
      * {schemeId} to a particular AuthenticationScheme class fully-specified name
      * Eg `authentication.scheme.basic.type = org.openmrs.module.authentication.web.BasicWebAuthenticationScheme`
@@ -84,7 +91,6 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String SCHEME_CONFIG_PREFIX_TEMPLATE = "authentication.scheme.{schemeId}.config.";
 
-    private static Properties config;
 
     private static final List<ClassLoader> classLoaders = new ArrayList<>();
 
@@ -232,6 +238,19 @@ public class AuthenticationConfig implements Serializable {
      */
     public static List<String> getWhiteList() {
         return getStringList(WHITE_LIST);
+    }
+
+     /**
+     * @return the List of url patterns to allow without force password authentication redirection
+     */
+    public static List<String> getPasswordChangeWhiteList() {
+        List<String> whiteList = getStringList(PASS_WORD_CHANGE_WHITE_LIST);
+        whiteList.add(getChangePasswordUrl());  // Add the change password URL to the whitelist
+        return whiteList;
+    }
+
+    public static String getChangePasswordUrl() {
+        return getConfig().getProperty(PASSWORD_CHANGE_URL, "/changePassword");
     }
 
 
--- a/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
+++ b/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
@@ -38,7 +38,7 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String SCHEME_ID = "{schemeId}";
 
-    public static final String REQUIRE_PASSWORD_CHANGE_ON_FIRST_LOGIN = "authentication.requirePasswordChangeOnFirstLogin";
+    public static final String SUPPORT_FORCED_PASSWORD_CHANGE = "authentication.supportForcedPasswordChange ";
 
     public static final String PASSWORD_CHANGE_URL = "authentication.passwordChangeUrl";
 
--- a/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
+++ b/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
@@ -33,8 +33,6 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String PREFIX = "authentication";
 
-    private static Properties config;
-
     /**
      * All configured authentication schemes are identified by a unique {schemeId} in the configuration
      */
@@ -90,6 +88,8 @@ public class AuthenticationConfig implements Serializable {
      * Eg `authentication.scheme.basic.config.loginPage = /module/authentication/basic.htm`
      */
     public static final String SCHEME_CONFIG_PREFIX_TEMPLATE = "authentication.scheme.{schemeId}.config.";
+
+    private static Properties config;
 
 
     private static final List<ClassLoader> classLoaders = new ArrayList<>();
--- a/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
+++ b/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
@@ -38,6 +38,10 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String SCHEME_ID = "{schemeId}";
 
+    public static final String REQUIRE_PASSWORD_CHANGE_ON_FIRST_LOGIN = "authentication.requirePasswordChangeOnFirstLogin";
+
+    public static final String PASSWORD_CHANGE_URL = "authentication.passwordChangeUrl";
+
     /**
      * This property determines which authentication scheme is used
      * This enables implementations to configure the preferred authentication scheme at runtime
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -19,6 +19,7 @@ import org.openmrs.module.authentication.AuthenticationCredentials;
 import org.openmrs.module.authentication.DelegatingAuthenticationScheme;
 import org.openmrs.module.authentication.UserLogin;
 import org.openmrs.module.authentication.UserLoginTracker;
+import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.web.WebConstants;
 import org.springframework.util.AntPathMatcher;
 
@@ -224,6 +225,11 @@ public class AuthenticationFilter implements Filter {
 		}
 		if (StringUtils.isNotBlank(redirect)) {
 			return contextualizeUrl(request, redirect);
+		}
+		// Check if password change on first login is required
+		if (AuthenticationConfig.getBoolean(AuthenticationConfig.REQUIRE_PASSWORD_CHANGE_ON_FIRST_LOGIN, false) &&
+				"true".equals(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD)) {
+			return contextualizeUrl(request, AuthenticationConfig.PASSWORD_CHANGE_URL);
 		}
 		return null;
 	}
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -184,11 +184,6 @@ public class AuthenticationFilter implements Filter {
 	 * @param request the request to use to determine url redirection
 	 */
 	protected String determineSuccessRedirectUrl(HttpServletRequest request) {
-		// Check if password change on first login is required
-		if (AuthenticationConfig.getBoolean(AuthenticationConfig.SUPPORT_FORCED_PASSWORD_CHANGE, false) &&
-		    AuthenticationConfig.getBoolean(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD, false)) {
-			return WebUtil.contextualizeUrl(request, AuthenticationConfig.PASSWORD_CHANGE_URL);
-		}
 		// First check for any "redirect" or "refererURL" parameters in the request, default to context path
 		String redirect = request.getParameter("redirect");
 		if (StringUtils.isBlank(redirect)) {
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -127,7 +127,7 @@ public class AuthenticationFilter implements Filter {
 
 					// If any credentials were passed in the request or session attempt to authentication with them
 					AuthenticationCredentials credentials = webScheme.getCredentials(session);
-					String challengeUrl = contextualizeUrl(request, webScheme.getChallengeUrl(session));
+					String challengeUrl = WebUtil.contextualizeUrl(request, webScheme.getChallengeUrl(session));
 					if (credentials != null) {
 						try {
 							session.removeErrorMessage();
@@ -147,7 +147,7 @@ public class AuthenticationFilter implements Filter {
 					}
 					// If no credentials were found, redirect to challenge url unless whitelisted
 					else {
-						if (!isWhiteListed(request)) {
+						if (!WebUtil.isWhiteListed(request, AuthenticationConfig.getWhiteList())) {
 							log.trace("Authentication required: " + request.getRequestURI());
 							session.sendRedirect(challengeUrl);
 						}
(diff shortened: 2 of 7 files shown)


omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -134,6 +134,11 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 							session.regenerateHttpSession();  // Guard against session fixation attacks
 							session.refreshDefaultLocale(); // Refresh context locale after authentication
 							String successUrl = determineSuccessRedirectUrl(request);
+							// Check if password change on first login is required
+                            if (AuthenticationConfig.isPasswordChangeOnFirstLoginRequired() && userLogin.isFirstLogin()) {
+                                 String passwordChangeUrl = request.getContextPath() + "/passwordChange.htm";
+                                  response.sendRedirect(passwordChangeUrl);
+                            }
```
ibacher, line 141: Fix the indentation here...

omod/src/test/java/org/openmrs/module/authentication/web/AuthenticationFilterTest.java
```diff
@@ -142,6 +143,29 @@ public void shouldRedirectToSuccessUrlIfAuthenticationSucceeds() throws Exceptio
...
+
+	    request.addParameter("redirect", "/passwordChange.htm");
+		request.addParameter("username", "admin");
+		request.addParameter("password", "adminPassword");
+        // Run the filter
+        filter.doFilter(request, response, chain);
+
+		assertThat(response.isCommitted(), equalTo(true));
+        
+        // Verify redirection to password change URL
```
ibacher, line 164: Indentation

api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
```diff
@@ -38,6 +38,12 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String SCHEME_ID = "{schemeId}";
 
+    public static final String REQUIRE_PASSWORD_CHANGE_ON_FIRST_LOGIN = "authentication.requirePasswordChangeOnFirstLogin";
+
+    public static boolean isPasswordChangeOnFirstLoginRequired() {
```
ibacher, line 43: Isn't this covered by `getBoolean()` below?

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -134,6 +134,11 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 							session.regenerateHttpSession();  // Guard against session fixation attacks
 							session.refreshDefaultLocale(); // Refresh context locale after authentication
 							String successUrl = determineSuccessRedirectUrl(request);
+							// Check if password change on first login is required
+                            if (AuthenticationConfig.isPasswordChangeOnFirstLoginRequired() && userLogin.isFirstLogin()) {
+                                 String passwordChangeUrl = request.getContextPath() + "/passwordChange.htm";
+                                  response.sendRedirect(passwordChangeUrl);
```
ibacher, line 140: So the page to redirect to _must_ be configurable. I would also plug this code into the `determineSuccessRedirectUrl()` function, so we only have a single part of code actually emitting a redirect.

api/src/main/java/org/openmrs/module/authentication/UserLogin.java
```diff
@@ -157,6 +158,20 @@ public synchronized void setUsername(String username) {
         this.username = username;
     }
 
+    /**
+     * @return true if it's the user's first login
+     */
+    public boolean isFirstLogin() {
+        return isFirstLogin;
+    }
```
ibacher, line 166: I don't see why we need these? The only thing we _should_ be checking is whether the _user_ has the user property `OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD` set to "true". That way we support both first login and additional password reset changes.

ibacher (changes requested):
Thanks @gitcliff! This is a reasonable first-pass, but see my comments. Also we should definitely have a ticket in the Authentication module Jira project for this. Could you please create one? I can fill in details.

gitcliff:
> Thanks @gitcliff! This is a reasonable first-pass, but see my comments. Also we should definitely have a ticket in the Authentication module Jira project for this. Could you please create one? I can fill in details.
thanks @ibacher for the review ,,,i have addressed the comments ..
I don't see the [JIRA board for this module ](https://openmrs.atlassian.net/jira/projects?page=1&sortKey=name&sortOrder=ASC) to file a ticket , should i create one ?

ibacher:
Created one here: https://openmrs.atlassian.net/jira/software/projects/AUT/

api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
```diff
@@ -38,6 +38,10 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String SCHEME_ID = "{schemeId}";
 
+    public static final String REQUIRE_PASSWORD_CHANGE_ON_FIRST_LOGIN = "authentication.requirePasswordChangeOnFirstLogin";
```
ibacher, line 41: Let's rename this to `SUPPORT_FORCED_PASSWORD_CHANGE`

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -225,6 +226,11 @@ protected String determineSuccessRedirectUrl(HttpServletRequest request) {
 		if (StringUtils.isNotBlank(redirect)) {
 			return contextualizeUrl(request, redirect);
 		}
+		// Check if password change on first login is required
+		if (AuthenticationConfig.getBoolean(AuthenticationConfig.REQUIRE_PASSWORD_CHANGE_ON_FIRST_LOGIN, false) &&
+				"true".equals(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD)) {
+			return contextualizeUrl(request, AuthenticationConfig.PASSWORD_CHANGE_URL);
+		}
```
ibacher, line 233: This actually needs to come _before_ the normal redirect processing

gitcliff: thanks @ibacher ,,,
kindly review

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -217,6 +218,11 @@ protected AuthenticationScheme getAuthenticationScheme() {
 	 * @param request the request to use to determine url redirection
 	 */
 	protected String determineSuccessRedirectUrl(HttpServletRequest request) {
+		// Check if password change on first login is required
+		if (AuthenticationConfig.getBoolean(AuthenticationConfig.SUPPORT_FORCED_PASSWORD_CHANGE, false) &&
+				"true".equals(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD)) {
+			return contextualizeUrl(request, AuthenticationConfig.PASSWORD_CHANGE_URL);
+		}
```
mseaton, line 225: @gitcliff I wouldn't think that this is something that we are just aiming to do once, as soon as the user is authenticated, and only then.  For example, if they authenticate, and we redirect them to this page, and then they just choose to enter a different url into the address bar, this wouldn't prevent them from doing that.  My assumption is that we want to force them to change their password without accessing any other pages.  So, this is really something we want to do independently of authentication.  I would think this could either be achieved via a separate servlet filter that runs after the AuthenticationFilter, or a separate check in the AuthenticationFilter that is placed at the end (eg. the code that is hit if the user is already authenticated).  Interested in @ibacher 's thoughts...

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -217,6 +218,11 @@ protected AuthenticationScheme getAuthenticationScheme() {
 	 * @param request the request to use to determine url redirection
 	 */
 	protected String determineSuccessRedirectUrl(HttpServletRequest request) {
+		// Check if password change on first login is required
+		if (AuthenticationConfig.getBoolean(AuthenticationConfig.SUPPORT_FORCED_PASSWORD_CHANGE, false) &&
+				"true".equals(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD)) {
```
mseaton, line 223: This is not actually checking the user property value, but is checking the value of the constant.  You actually need to fetch the user property value with this name, not just compare the constant itself.

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
ibacher: Hmmmm... That's actually a good point. Maybe having this as a second filter makes the most sense. There will need to be some additional configuration around this too... I.e., we need to white-list certain URLs (e.g., to allow the user to hit the "change password" endpoint).

gitcliff: thanks @ibacher  and @mseaton

api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
```diff
@@ -70,6 +72,11 @@ public class AuthenticationConfig implements Serializable {
      */
     public static final String WHITE_LIST = "authentication.whiteList";
 
+    /**
+     * URLs that might need to be made accessible without go through the password change authentication.
+     */
+    public static final String PASS_WORD_CHANGE_WHITE_LIST = "authentication.passWordChangeWhiteList";
```
mseaton, line 78: `PASS_WORD` -> `PASSWORD`, and `passWord` -> `password`

api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
```diff
@@ -234,6 +240,19 @@ public static List<String> getWhiteList() {
...
+     * @return the List of url patterns to allow without force password authentication redirection
+     */
+    public static List<String> getPasswordChangeWhiteList() {
+        List<String> whiteList = getStringList(PASS_WORD_CHANGE_WHITE_LIST);
+        whiteList.add(getChangePasswordUrl());  // Add the change password URL to the whitelist
+        return whiteList;
+    }
+
+    public static String getChangePasswordUrl() {
+        return getConfig().getProperty(PASSWORD_CHANGE_URL, "/changePassword");
```
mseaton, line 253: Does this refer to an actual /changePassword page somewhere?  If we are going to have a default, we should make sure it actually works under some standard conditions (eg. with legacyui running)

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import org.openmrs.module.authentication.AuthenticationConfig;
+import org.openmrs.module.authentication.UserLogin;
+import org.openmrs.util.OpenmrsConstants;
+
+/**
+ * This filter checks if an authenticated user has been flagged by the admin to change his password
+ * on first/subsequent login. It will intercept any requests made to a *.html or a *.form to force
+ * the user to change his password.
```
mseaton, line 22: Why just a *.html and *.form?  I don't think this comment is accurate, is it?

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class ForcePasswordChangeFilter implements Filter {
+	
+	private String excludeURL;
+	
+	private String changePasswordForm;
```
mseaton, line 28: This doesn't seem right.  Why do you have this here?  Aren't you getting this from the AuthenticationConfig, which is able to be specified at runtime, not a hard-coded filter param?  Remove this.

mseaton (changes requested):

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
gitcliff: thanks  @mseaton
 had skipped removing it but removed now

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -217,33 +184,23 @@ protected AuthenticationScheme getAuthenticationScheme() {
 	 * @param request the request to use to determine url redirection
 	 */
 	protected String determineSuccessRedirectUrl(HttpServletRequest request) {
+		// Check if password change on first login is required
+		if (AuthenticationConfig.getBoolean(AuthenticationConfig.SUPPORT_FORCED_PASSWORD_CHANGE, false) &&
+		    AuthenticationConfig.getBoolean(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD, false)) {
```
mseaton, line 189: This is not correct.  The USER_PROPERTY_CHANGE_PASSWORD is not coming from the AuthenticationConfig, but from the user properties.

mseaton: Actually, sorry - this entire set of changes should not be here at all.  This is now covered by the other filter.  You need to remove these changes to this method.

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,80 @@
...
+		HttpServletResponse response = (HttpServletResponse) servletResponse;
+
+		if (WebUtil.isWhiteListed(request, AuthenticationConfig.getPasswordChangeWhiteList())) {
+			chain.doFilter(request, response);
+			return;
+		}
+		AuthenticationSession session = getAuthenticationSession(request, response);
+		UserLogin userLogin = session.getUserLogin();
+		String changePasswordProperty = userLogin.getUser().getUserProperties()
+				.get(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD);
```
mseaton, line 49: You should be able to use `userLogin.getUser().getUserProperty(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD)`

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	public void destroy() {
+	}
+
+	@Override
+	public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain chain)
+			throws IOException, ServletException {
+		HttpServletRequest request = (HttpServletRequest) servletRequest;
+		HttpServletResponse response = (HttpServletResponse) servletResponse;
+
```
mseaton, line 41: Previously, you had a check for `(AuthenticationConfig.getBoolean(AuthenticationConfig.SUPPORT_FORCED_PASSWORD_CHANGE, false)`.  Are you no longer supporting this?  I would expect to see that in this filter.  And I'd also recommend checking this once and saving it so that you can make this filter as efficient as possible for every request.

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,80 @@
...
+		if (WebUtil.isWhiteListed(request, AuthenticationConfig.getPasswordChangeWhiteList())) {
+			chain.doFilter(request, response);
+			return;
+		}
+		AuthenticationSession session = getAuthenticationSession(request, response);
+		UserLogin userLogin = session.getUserLogin();
+		String changePasswordProperty = userLogin.getUser().getUserProperties()
+				.get(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD);
+		Boolean changePasswordFlag = AuthenticationConfig.getBoolean(changePasswordProperty, false);
```
mseaton, line 50: This is not correct.  You don't use the AuthenticationConfig to do this, you would use something like `BooleanUtils.isTrue()`

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,80 @@
...
+		if (userLogin.isUserAuthenticated() && BooleanUtils.isTrue(changePasswordFlag)) {
+			String changePasswordUrl = AuthenticationConfig.getChangePasswordUrl();
```
mseaton, line 52: This seems like something that should be evaluated once and cached, not evaluated and checked per request.

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
gitcliff: thanks @mseaton 
i have added the check and cached it

api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java
```diff
@@ -82,6 +91,7 @@ public class AuthenticationConfig implements Serializable {
 
     private static Properties config;
 
+
```
mseaton, line 94: Remove this newline

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,91 @@
...
+	String changePasswordUrl;
+
+
+	protected final Log log = LogFactory.getLog(getClass());
+
+	@Override
+	public void init(FilterConfig config) throws ServletException {
+		this.config = config;
+		this.changePasswordUrl = AuthenticationConfig.getChangePasswordUrl();
+		this.supportForcedPasswordChange = BooleanUtils.isTrue(AuthenticationConfig.getBoolean
```
mseaton, line 40: The BooleanUtils.isTrue here is unnecessary.  Just use the AuthenticationConfig method

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	@Override
+	public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain chain)
+			throws IOException, ServletException {
+		HttpServletRequest request = (HttpServletRequest) servletRequest;
+		HttpServletResponse response = (HttpServletResponse) servletResponse;
+
+		if (WebUtil.isWhiteListed(request, AuthenticationConfig.getPasswordChangeWhiteList()) 
+		    || !supportForcedPasswordChange) {
```
mseaton, line 56: I would evaluate `supportForcedPasswordChange` first, as it is faster and if it is true then it won't bother evaluating the next.  Why aren't you also caching the whitelist value at startup?

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,91 @@
...
+
+	/**
+	 * Return a valid AuthenticationSession for the given request
+	 * 
+	 * @param request the HttpServletRequest to use to retrieve the
+	 *                AuthenticationSession
+	 * @return the AuthenticationSession associated with this HttpServletRequest
+	 */
+	protected AuthenticationSession getAuthenticationSession(HttpServletRequest request, 
+	  HttpServletResponse response) {
```
mseaton, line 87: Ugly

mseaton (changes requested):
Looking better.  Please re-request review via the github PR tools when you are ready for this to be looked at again.

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
gitcliff: yeah, thanks @mseaton ,, cached the whitelist value at startup

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -0,0 +1,92 @@
...
+		if (!supportForcedPasswordChange || WebUtil.isWhiteListed(request, whiteList)) {
+			chain.doFilter(request, response);
+			return;
+		}
+
+		AuthenticationSession session = getAuthenticationSession(request, response);
+		UserLogin userLogin = session.getUserLogin();
+		String changePasswordProperty = userLogin.getUser().
+		  getUserProperty(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD);
+		Boolean changePasswordFlag = BooleanUtils.isTrue(Boolean.valueOf(changePasswordProperty));
```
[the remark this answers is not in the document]
gitcliff: thanks @ibacher fixed it

omod/src/main/java/org/openmrs/module/authentication/web/ForcePasswordChangeFilter.java
```diff
@@ -64,14 +67,9 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 		UserLogin userLogin = session.getUserLogin();
 		String changePasswordProperty = userLogin.getUser().
 		  getUserProperty(OpenmrsConstants.USER_PROPERTY_CHANGE_PASSWORD);
-		Boolean changePasswordFlag = BooleanUtils.isTrue(Boolean.valueOf(changePasswordProperty));
+		boolean changePasswordFlag = Boolean.parseBoolean(changePasswordProperty);
 		if (userLogin.isUserAuthenticated() && BooleanUtils.isTrue(changePasswordFlag)) {
```
ibacher, line 71: So, as I noted, you can get rid of the `BooleanUtils.isTrue()` call on this line too.

ibacher (approved):

mseaton (approved):

PR #29, opened 2026-03-20
Feat(filter): inject Location header for all unauthenticated REST calls, not just/session

shubhangiisinghh:
## SUMMARY

When authentication is incomplete, `AuthenticationFilter` previously only injected a
`Location` header for `/ws/rest/*/session`. Any other REST call (e.g. `/ws/rest/v1/patient`)
received a bare 302 redirect to the JSP `challengeUrl`, which O3 SPA clients cannot render.

This PR broadens the pattern to `/ws/rest/**`:

- **Session endpoint**-`Location` header injected, request passes through. The REST module
  writes its own JSON body; `openmrs-esm-core` reads the `Location` header on every session
  poll to detect an auth challenge.
- **All other REST endpoints**- `Location` header injected + 401 via `handleAuthenticationFailure`,
  which already respects the `nonRedirectUrls` config. SPA clients receive a machine-readable
  signal to redirect the user to login.
- **Non-REST requests** -completely unaffected (JSP pages, static assets, whitelisted URLs).

No new dependencies. No JSON duplication. Backward compatible with 2.x deployments.

## Approach

This directly implements the design discussed in the [PR #18 review](https://github.com/openmrs/openmrs-module-authentication/pull/18)
by @ibacher and @mseaton: let the session endpoint write its own body rather than
replicating it in the filter, and avoid taking a dependency on `webservices.rest`.

## Testing

Three new unit tests in `AuthenticationFilterTest` cover the three distinct behaviors:
1. Session endpoint gets `Location` header and request proceeds (no 401)
2. Other REST endpoints get `Location` header + 401
3. Non-REST endpoints still get 302 (unchanged behavior)

### DIFF
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -148,16 +148,23 @@ public class AuthenticationFilter implements Filter {
 						}
 					}
 					// If no credentials were found, redirect to challenge url unless whitelisted
-					else {
-						if (WebUtil.matchesPath(request, "/ws/rest/*/session")) {
-							// Add a location header to the session endpoint to support frontend redirection to login
-							response.setHeader("Location", challengeUrl);
-						}
-						else if (!WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getWhiteList())) {
-							log.trace("Authentication required: " + request.getRequestURI());
-							handleAuthenticationFailure(request, response, challengeUrl);
-						}
-					}
+else {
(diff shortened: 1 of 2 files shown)


shubhangiisinghh:
@jayasanka-sack @ibacher Would love any feedback, especially on whether this aligns with the direction you have in mind for O3 support.

jayasanka-sack:
Thanks for the PR @shubhangiisinghh.

shubhangiisinghh:
Thanks for the context @jayasanka-sack! I'll make sure to include this in my proposal. 
Hoping to work on it properly during GSoC if I get a chance !

PR #31, opened 2026-03-22
test(config): add missing tests for getNonRedirectUrls, getChangePasswordUrl, and getPasswordChangeWhiteList

shubhangiisinghh:
## What

While reading through `AuthenticationConfig` to understand the force-password-change flow, I notced that three methids had no test coverage at all:

- `getNonRedirectUrls()`- returns the hardcoded `/ws/**/*` pattern plus anything configured via `authentication.nonRedirectUrls`
- `getChangePasswordUrl()`-reads `authentication.passwordChangeUrl`
- `getPasswordChangeWhiteList()`- builds on the standard whitelist and adds the change-password URL itself so the page can render

Every other method in `AuthenticationConfig` already has tests. These three were just missed. This adds them to the existing `AuthenticationConfigTest` in the same style as the rest of the file.

## What this is not

No logic changes. No new files, Just filling the gap.

### DIFF
--- a/api/src/test/java/org/openmrs/module/authentication/AuthenticationConfigTest.java
+++ b/api/src/test/java/org/openmrs/module/authentication/AuthenticationConfigTest.java
@@ -1,10 +1,5 @@
 package org.openmrs.module.authentication;
 
-import org.junit.jupiter.api.Test;
-import org.openmrs.User;
-import org.openmrs.api.context.AuthenticationScheme;
-import org.openmrs.api.context.UsernamePasswordAuthenticationScheme;
-
 import java.util.Date;
 import java.util.List;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


shubhangiisinghh:
@jayasanka-sack @ibacher looking forward to any feedback when you get a chance!

api/src/test/java/org/openmrs/module/authentication/AuthenticationConfigTest.java
```diff
@@ -114,7 +115,34 @@ public void shouldGetWhiteList() {
 		assertThat(patterns.size(), equalTo(21));
 		assertThat(patterns.contains("*.pdf"), equalTo(true));
 	}
+    @Test
+public void shouldGetNonRedirectUrls() {
```
dulajweligaththa08, line 119: Good addition of test cases. This improves coverage for authentication configuration and helps ensure expected behavior.

PR #32, opened 2026-03-22
ci: update release workflow to use OMRS_BOT_GH_TOKEN secret

shubhangiisinghh:
PR #27 updated `build.yml` to use `BOT_GH_TOKEN: ${{ secrets.OMRS_BOT_GH_TOKEN }}`
after the shared reusable workflow updated its expected secret name.

`release.yml` still had the old `BOT_PAT: ${{ secrets.BOT_PAT }}` reference
and wasn't updated at the same time. This brings it in line with `build.yml`.
```


jayasanka-sack:
Thanks for the PR, @shubhangiisinghh! I appreciate you taking the time to work on this.

The BOT_GH_TOKEN is only needed in the build.yml file for pushing the OWASP report. You can find more details here: https://openmrs.atlassian.net/browse/O3-5519

I’m going to go ahead and close this PR for now but thanks again for the contribution!

PR #33, opened 2026-03-26
(fix) AUT-6: Normalize TOTP code by removing whitespace before verification

NitinKumar1-1:
## Description
Fixes TOTP verification failure when codes contain whitespace.

Some authenticator apps (e.g., Microsoft Authenticator) display codes with spaces (e.g., "123 456"),
but verification expects a continuous string.

## Solution
Normalize the input code by removing all whitespace before verification using StringUtils.deleteWhitespace.

## Testing
- Verified that codes with and without spaces are accepted

## Related Issue
AUT-6

### DIFF
--- a/omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
@@ -104,7 +104,9 @@ public class TotpAuthenticationScheme extends WebAuthenticationScheme {
 			throw new ContextAuthenticationException("authentication.error.noSecretConfiguredForUser");
 		}
 		String decodedSecret = Security.decrypt(userSecret);
-		if (!verifyCode(decodedSecret, c.code)) {
+		// Normalize code to remove whitespace (some authenticator apps include spaces)
+		String normalizedCode = StringUtils.deleteWhitespace(c.code);
+		if (!verifyCode(decodedSecret, normalizedCode)) {
 			throw new ContextAuthenticationException("authentication.error.invalidCredentials");
 		}
 


NitinKumar1-1:
Could a maintainer please approve the workflow so that checks can run?

Happy to make any changes if needed. Thanks!

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -104,7 +104,9 @@ protected Authenticated authenticate(AuthenticationCredentials credentials, User
 			throw new ContextAuthenticationException("authentication.error.noSecretConfiguredForUser");
 		}
 		String decodedSecret = Security.decrypt(userSecret);
-		if (!verifyCode(decodedSecret, c.code)) {
+		// Normalize code to remove whitespace (some authenticator apps include spaces)
+		String normalizedCode = StringUtils.deleteWhitespace(c.code);
+		if (!verifyCode(decodedSecret, normalizedCode)) {
 			throw new ContextAuthenticationException("authentication.error.invalidCredentials");
 		}
```
dulajweligaththa08, line 112: Good improvement! Normalizing the code by removing whitespace helps handle inputs from authenticator apps more reliably.

NitinKumar1-1: @dulajweligaththa08 Thanks for the review! Glad this improvement helps.
Please let me know if any further changes are needed .

kamakhyadhruv5 (commented):
Hi @NitinKumar1-1 👋

I reviewed your PR for normalizing the TOTP code before verification. Overall, this is a good improvement and addresses a real-world issue where authenticator apps may include spaces in the code.

### ✅ What looks good:

* Good catch on handling whitespace in TOTP codes 👍
* Use of `StringUtils.deleteWhitespace()` is clean and effective
* The added comment clearly explains the reason for normalization

### 🔍 Suggestions:

* Instead of creating a new variable `normalizedCode`, you could consider normalizing `c.code` once at the beginning to avoid multiple usages of the raw value
* It might be helpful to add a null check for `c.code` before processing to prevent any unexpected issues
* Consider adding a small unit test to cover cases where the input contains spaces (e.g., `"123 456"`)

### ❓ Question:

* Should we also handle other possible formatting issues (like leading/trailing spaces or non-numeric characters), or is whitespace normalization sufficient for now?

### 💡 Overall:

This is a meaningful and clean fix. With minor improvements and test coverage, it will be even more robust. Good work 👍

shubhangiisinghh:
Hey @NitinKumar1-1 , 
One small thought- should the whitespace stripping happen in `getCredentials()` when the code is first read from the request, rather than later in the flow? That way the raw code with spaces is never stored in `TotpCredentials` at all, which feels slightly cleaner. Just wondering if that placement would be more intentional.

jayasanka-sack (changes requested):
Thanks for the PR @NitinKumar1-1. A simple `trim()` would be sufficient here. Handling other formatting variations is generally a frontend concern, so I don't think we need additional whitespace normalization in the backend.

Could you please update the PR accordingly and let us know once it's ready for another review?

dilankavishka:
@jayasanka-sack  I think we can close this PR as this was fixed by https://github.com/openmrs/openmrs-module-authentication/pull/55.

jayasanka-sack:
Thanks for working on this @NitinKumar1-1!  This has now been addressed and merged in #55, so I’m closing this PR.

PR #37, opened 2026-03-28
Prototype SPA-friendly authentication failure response for O3 integration

sudeshkar:
## Summary

This PR updates `AuthenticationFilter` to return a `401 + Location` 
header for all SPA/API clients when authentication is incomplete, 
rather than a 3xx redirect.

## Problem

Previously, the `Location` header was only set for `/ws/rest/*/session` 
requests. Any other REST call made while authentication was incomplete 
would receive a 3xx redirect that the O3 SPA cannot follow.

## Changes

- Extended `Location` header behavior to all URLs matching 
  `authentication.nonRedirectUrls` — not just the session endpoint
- Removed `resolveChallengeType()` and `resolveSpaRedirectUrl()` 
  methods — auth state must not be exposed in the response body
- `handleAuthenticationFailure()` now returns `401 + Location` 
  for all non-redirect URLs

## Behavior

| Client | Response |
|--------|----------|
| SPA/API (nonRedirectUrl) | `401 + Location` header → frontend navigates |
| Browser | `3xx redirect` → existing behavior unchanged |

## Notes

- No auth state is exposed in the response body
- The backend simply tells the frontend "go to this page" via 
  the Location header — that the backend 
  should not reveal which auth step is pending
- No other backend changes required
- Fully backward compatible — existing 2.x deployments unaffected

Ref: https://talk.openmrs.org/t/gsoc-2026-o3-frontend-architecture-for-authentication-module-integration/49131

### DIFF
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -152,6 +152,7 @@ public class AuthenticationFilter implements Filter {
 						if (WebUtil.urlMatchesAnyPattern(request, 
 							AuthenticationConfig.getNonRedirectUrls())) {
 							response.setHeader("Location", challengeUrl);
+							response.sendError(HttpServletResponse.SC_UNAUTHORIZED);
 						}
 						else if (!WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getWhiteList())) {
 							log.trace("Authentication required: " + request.getRequestURI());
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -175,42 +175,20 @@ public class AuthenticationFilter implements Filter {
 	 * @param request the request to handle
 	 * @param challengeUrl the challengeUrl to direct the response to
 	 */
-	protected void handleAuthenticationFailure(
-        HttpServletRequest request,
-        HttpServletResponse response,
-        String challengeUrl) throws IOException {
+	protected void handleAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, String challengeUrl) throws IOException {
 
-    if (WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getNonRedirectUrls())) {
+    boolean isNonRedirectUrl = WebUtil.urlMatchesAnyPattern(
+        request,
+        AuthenticationConfig.getNonRedirectUrls()
+    );
 
+    if (isNonRedirectUrl) {
         response.setHeader("Location", challengeUrl);
-
-        // Detect REST / SPA request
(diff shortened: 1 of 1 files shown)


omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -175,16 +175,87 @@ else if (!WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getWhiteLis
...
-	protected void handleAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, String challengeUrl) throws IOException {
-		if (WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getNonRedirectUrls())) {
-			response.setHeader("Location", challengeUrl);
-			response.sendError(HttpServletResponse.SC_UNAUTHORIZED);
-		}
-		else {
-			response.sendRedirect(challengeUrl);
-		}
-	}
-
```
dulajweligaththa08, line 187: Maybe we can improve this part for better readability.

sudeshkar: Thanks for the suggestion! I refactored the `handleAuthenticationFailure` method to improve readability by introducing a descriptive variable and reducing nesting. Please let me know if further adjustments would help.

shubhangiisinghh:
Hey @sudeshkar ,
Interesting approach! One thing worth considering -@ibacher mentioned on the Talk thread that exposing auth state details in the response body could be a security concern (https://talk.openmrs.org/t/seeking-guidance-integrate-o3-with-the-authentication-module/48857). 
The `Location` header approach might be safer here since it tells the frontend where to go without revealing which auth step is pending. Curious what the maintainers think about the tradeoff.

sudeshkar:
> Hey @sudeshkar , Interesting approach! One thing worth considering -@ibacher mentioned on the Talk thread that exposing auth state details in the response body could be a security concern (https://talk.openmrs.org/t/seeking-guidance-integrate-o3-with-the-authentication-module/48857). The `Location` header approach might be safer here since it tells the frontend where to go without revealing which auth step is pending. Curious what the maintainers think about the tradeoff.

Thanks [@shubhangiisinghh](https://github.com/shubhangiisinghh) — that's a valid point and you're absolutely right. I've updated the PR to remove the JSON body approach entirely. The updated implementation now returns only a 401 + Location header for SPA/API clients, with no auth state exposed in the response body — aligning with @ibacher's guidance that the backend should simply tell the frontend where to go without revealing which auth step is pending. Appreciate you flagging this!

jayasanka-sack:
Hi @sudeshkar, thank you for the PR. This scenario is already being handled by the existing logic here:

https://github.com/openmrs/openmrs-module-authentication/blob/58592bba56f037262a0dcfe633a93831a4a4fa21/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java#L179

Really appreciate you taking the initiative to investigate and submit a fix.

PR #45, opened 2026-04-07
AUT-20 - Support code via email authentication scheme

mseaton:


mogoodrich (approved):
LGTM, and happy to play around with this and test this out... hopefully I'm not showing my naivete  in how you set up this whole authentication pattern, (and how email authentication works) but this ends up storing the email code in the session, so it doesn't have to go in the database anywhere?  (not saying that's a problem, just makes me realize I didn't know/think about how email authentication generally worked).

mseaton:
> LGTM, and happy to play around with this and test this out... hopefully I'm not showing my naivete in how you set up this whole authentication pattern, (and how email authentication works) but this ends up storing the email code in the session, so it doesn't have to go in the database anywhere? (not saying that's a problem, just makes me realize I didn't know/think about how email authentication generally worked).

Marking this for re-review @mogoodrich since I've added some things.  Basically this adds the Email authentication scheme, but it also modifies the workings of the 2FA authentication scheme such that the user can have multiple valid 2FA methods enabled on their account, the first of which is the default but any of which can be set as the currently selected one for the current session.

The idea is that the user:
* Logs in with first factor
* Is presented with their default (first ordered) second factor method, and links to change to a different method
* Can provide their second factor on the current (default) page or in any of the other linked auth pages, and all would be sufficient for 2FA.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
```diff
@@ -0,0 +1,235 @@
...
+		emailSubject = config.getProperty(EMAIL_SUBJECT, "authentication.email.subject");
+		emailFrom = config.getProperty(EMAIL_FROM, "");
+		resendParam = config.getProperty(RESEND_PARAM, "resend");
+	}
+
+	/**
+	 * @return the name of the user property that stores the verified email address that a given code should be sent
+	 */
+	public String getVerifiedEmailUserPropertyName() {
+		return "authentication." + getSchemeId() + ".verifiedEmail";
```
mseaton, line 78: Rather than always use the user.getEmail, I am explicitly saving an email user property that is for this authentication scheme, as this gives more control over validating that it has been verified.  Otherwise, there isn't any good way to determine whether the user.getEmail was arbitrarily updated and is validated for use for authentication

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
```diff
@@ -145,9 +146,15 @@ public Authenticated authenticate(AuthenticationCredentials credentials, UserLog
 		if (!mfaCreds.validatedCredentials.contains(getPrimaryAuthenticationScheme().getSchemeId())) {
 			throw new ContextAuthenticationException("authentication.error.primaryAuthenticationRequired");
 		}
-		WebAuthenticationScheme secondaryScheme = getSecondaryAuthenticationScheme(mfaCreds.user);
-		if (secondaryScheme != null) {
-			if (!mfaCreds.validatedCredentials.contains(secondaryScheme.getSchemeId())) {
```
mseaton, line 150: All of the changes in this class are actually unrelated to this feature, but are related to another feature that I decided to work on at the same time, which is to allow the user to have multiple active/enabled methods of 2FA that they can use and choose from if needed.  So they can have a primary second factor of TOTP authenticator app, but also have one using Email that they can choose as an alternative if needed (eg. if they accidentally deleted their configuration from their app / got a new phone, or if their phone is dead, etc).  Similarly, if they are disconnected from internet and can't get an email, but it is their primary method, they could choose to use authenticator app.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
```diff
@@ -182,29 +189,74 @@ protected WebAuthenticationScheme getPrimaryAuthenticationScheme() {
...
 	 * @return the WebAuthenticationScheme to use for the given MultiFactorAuthenticationCredentials
 	 */
-	protected WebAuthenticationScheme getSecondaryAuthenticationScheme(User user) {
+	public WebAuthenticationScheme getSecondaryAuthenticationScheme(AuthenticationSession session, User user) {
 		if (user != null) {
-			String secondaryName = user.getUserProperty(USER_PROPERTY_SECONDARY_TYPE);
-			if (StringUtils.isNotBlank(secondaryName)) {
-				AuthenticationScheme scheme = AuthenticationConfig.getAuthenticationScheme(secondaryName);
+			String preferredScheme = null;
+			Object sessionVal = session.getHttpSessionAttributes().get(getSessionKeyForSecondarySchemeId());
```
mseaton, line 217: Basically, I have changed the user property from a single schemeId value to a comma-separated list of schemeIds, the first of which would be used / presented by default, but any of which could be selected for use (assuming the UI has this feature built).  Selecting a different one involves setting the scheme id desired as a particular session attribute.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
```diff
@@ -0,0 +1,235 @@
...
+		if (StringUtils.isBlank(c.submittedCode)) {
+			throw new ContextAuthenticationException("authentication.error.codeRequired");
+		}
+		if (userLogin.getUser() != null && !userLogin.getUser().equals(c.user)) {
+			throw new ContextAuthenticationException("authentication.error.userDiffersFromCandidateUser");
+		}
+		if (c.expectedExpiry == null || System.currentTimeMillis() > c.expectedExpiry) {
+			throw new ContextAuthenticationException("authentication.error.codeExpired");
+		}
+		if (!c.submittedCode.equals(c.expectedCode)) {
```
ibacher: So this is one thing that falls into why people usually don't advise building your own authentication schemes, but this code has a subtle issues that make this mechanism of authentication vulnerable to attack.

`String.equals()` actually short-circuits at the first point `c.submittedCode` does not equal `c.expectedCode()`. This enables what's called a timing attack where an attacker can start to glean information about the expected code by submitting multiple codes. E.g., if the expected code was "ABCDEF" and the attacker submits "AAAAAA" and "ABBBBB" the slight difference in timing between the reject of the first string (at the second character) and the second string (at the third character) can allow the attacker to infer where the mismatch occurs. To fix this, we need to use a comparison method that is guaranteed to be constant-time like `!MessageDigest.isEqual(c.submittedCode.getBytes(), c.expectedCode.getBytes());` (`MessageDigest.isEqual()` is designed to run in constant time as long as the `a` and `b` codes are the same length).

Relatedly, there's no rate limiting on:

1. How much time can pass between the user submitting two different MFA codes.
2. How frequently a user can request to be resent a code.

Which make it slightly easier for an attacker to take advantage of the timing attack.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
```diff
@@ -0,0 +1,235 @@
...
+	 */
+	protected void sendCode(User user, String code) {
+		String email = getVerifiedEmailForUser(user);
+		if (StringUtils.isBlank(email)) {
+			throw new ContextAuthenticationException("authentication.error.noEmailConfiguredForUser");
+		}
+		try {
+			Context.addProxyPrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES);
+			MessageSource messageSource = Context.getMessageSourceService().getActiveMessageSource();
+			String subject = messageSource.getMessage(emailSubject, new Object[] {code}, Context.getLocale());
```
ibacher, line 194: While it's not that I haven't seen people do this, it's generally a bad idea to put the MFA token in the email subject.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
```diff
@@ -69,9 +70,9 @@ public String getChallengeUrl(AuthenticationSession session) {
 		if (candidateUser == null) {
 			return getPrimaryAuthenticationScheme().getChallengeUrl(session);
 		}
-		WebAuthenticationScheme secondaryScheme = getSecondaryAuthenticationScheme(candidateUser);
+		WebAuthenticationScheme secondaryScheme = getSecondaryAuthenticationScheme(session, candidateUser);
 		if (secondaryScheme != null) {
-			return getSecondaryAuthenticationScheme(candidateUser).getChallengeUrl(session);
+			return getSecondaryAuthenticationScheme(session, candidateUser).getChallengeUrl(session);
```
ibacher: nit: but is there a reason to call `getSecondaryAuthenticationScheme()` twice here?

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
```diff
@@ -182,29 +189,74 @@ protected WebAuthenticationScheme getPrimaryAuthenticationScheme() {
...
+
 	/**
 	 * This returns the WebAuthenticationScheme that is configured for the given User
 	 * This is configured via a user property named `authentication.secondaryType`
 	 * This throws an Exception the configured AuthenticationScheme is not a WebAuthenticationScheme
 	 * @param user the User to check for configured secondary WebAuthenticationScheme
 	 * @return the WebAuthenticationScheme to use for the given MultiFactorAuthenticationCredentials
 	 */
-	protected WebAuthenticationScheme getSecondaryAuthenticationScheme(User user) {
+	public WebAuthenticationScheme getSecondaryAuthenticationScheme(AuthenticationSession session, User user) {
```
ibacher, line 214: Is there a reason for the protected -> public change here?

omod/src/test/java/org/openmrs/module/authentication/web/EmailAuthenticationSchemeTest.java
```diff
@@ -0,0 +1,280 @@
...
+		assertThat(authenticated.getAuthenticationScheme(), equalTo("email"));
+		assertThat(authenticated.getUser().getUsername(), equalTo("testing"));
+	}
+
+	@Test
+	public void shouldFailToAuthenticateWithIncorrectCode() {
+		getCredentials(null, null);
+		AuthenticationCredentials credentials = getCredentials("000000", null);
+		// Only fails if 000000 isn't the actual sent code (extremely unlikely with 6 digits)
+		if (!authenticationScheme.getLastSentCode().equals("000000")) {
```
ibacher: Since we're just using strings at the end of the day, couldn't this just be "ABCDEF" or something similar that the code could never be?

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	/**
+	 * @return a randomly generated numeric code of the configured length
+	 */
+	protected String generateCode() {
+		int max = (int) Math.pow(10, codeLength);
+		int code = new SecureRandom().nextInt(max);
+		return String.format("%0" + codeLength + "d", code);
+	}
```
ibacher: Since we seem to target 2.7.0+, this could instead be:

```java
RandomStringUtils.secureStrong().next(codeLength, "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789");
```

That ups the code entropy to ~31 bits from ~20 bits or, more concretely, gives you 1,402,410,240 possible codes instead of 151,200 without really increasing the difficulty in entering it.

omod/src/test/java/org/openmrs/module/authentication/web/EmailAuthenticationSchemeTest.java
```diff
@@ -0,0 +1,280 @@
...
+
+	// Configuration tests
+
+	@Test
+	public void shouldConfigureFromRuntimeProperties() {
+		assertThat(authenticationScheme.getSchemeId(), equalTo("email"));
+		assertThat(authenticationScheme.getChallengeUrl(authenticationSession), equalTo("/emailCode.page"));
+	}
+
+	// isUserConfigurationRequired tests
```
ibacher, line 96: Some of the tests in this section seem weird since `isUserConfigurationRequired()` only actually checks the verified email property.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
ibacher: And, yeah, this stuff is also broken for passwords in core. I'm hoping to fix that with GSoC this year.

mseaton: Interesting points, thanks for the detailed write up on this @ibacher !

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
mseaton: This is really a personal preference, which I have seen plenty of times in the wild, so figured it passed muster, but if we feel it is not best practice, I can change it.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
mseaton: Sounds good to me.  TBH, I was experimenting with this class by using Claude Code to see how it would do it writing it, so any feedback you have is interesting from that perspective.  Not sure if that has any impact on how much you trust it...one way or the other.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
mseaton: Yes, this allows one to query the currently selected secondary authentication scheme for the user, which is used, for example, by the authenticationui module to present the alternative second factor schemes one might switch to (by excluding the currently selected from the list of alternative options)

omod/src/test/java/org/openmrs/module/authentication/web/EmailAuthenticationSchemeTest.java
mseaton: Yeah, this class was pretty much entirely Claude Code generated when I had it write the scheme.  Probably needs more review and refinement.  I'll do that.

ibacher: To be fair, one of the things I actually quite like LLMs for is writing tests. [This](https://github.com/openmrs/openmrs-esm-core/pull/1735) recent PR I did relied pretty heavily on Claude for generating the tests and while the tests are mostly rote, it did turn up one [bug I was surprised no one pointed out](https://github.com/openmrs/openmrs-esm-core/pull/1735/changes#diff-abe791e8a18df243a35b3446abcaa6d4e8e9ebd734cd18c1634c4058da6e836aR142) (where the O3 dev tools wouldn't actually properly work if you were using a URL other than `/openmrs/spa` to serve content from). It does result in some things I never would've written (like [this construct](https://github.com/openmrs/openmrs-esm-core/pull/1735/changes#diff-63d128b2c1eb1c62a53db1e43353faa980f8161de4e9f4160a8ef5d607a0b4f4R18)), but it works.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
mseaton: OK, so I have modified things so that there is now a configurable email subject _and_ a configurable email body, and the subject defaults to "Your verification code" and the body defaults to "Your verification code is XXXXX".  An implementation can change this if they wish, but the default matches your recommendation.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
```diff
@@ -0,0 +1,241 @@
...
+		finally {
+			Context.removeProxyPrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES);
+		}
+	}
+
+	/**
+	 * @return a randomly generated numeric code of the configured length
+	 */
+	public String generateCode() {
+		return RandomStringUtils.secureStrong().next(codeLength, codeCharacters);
```
mseaton, line 211: I have changed this to use RandomStringUtils as you recommended @ibacher , but after giving this a bit more thought, I'm not sure about changing the character set to include all alphanumerics by default.  For example, for the same reason we use a Base 30 character set for identifier generation, there are some characters (eg 0 and O) that may be difficult to distinguish, and I'd rather it be unambiguous.  So what I have done is made this configurable, with a default as 0-9.  If an implementation wanted this to be more random, they could add characters to it, but at least from personal experience, almost all of the texts and emails I receive with 2FA codes are 6-digit numerics, so I'm going to stick with that as the default unless we feel strongly otherwise.

omod/src/test/java/org/openmrs/module/authentication/web/EmailAuthenticationSchemeTest.java
mseaton: 100%.  The issue here was Claude wrote tests for it's initial implementation.  Then I changed the implementation but neglected to update the tests, and the tests were not specific enough to start failing after my change.

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
mseaton: I have made the adjustment to use the MessageDigest.isEquals method for the reasons you suggest @ibacher .  I will hold off on adding any constraints on the frequency of sending new codes - we can ticket that for the future if we feel it is an important thing to add in.  We don't do anything like that for TOTP either, as far as I know.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
mseaton: Fixed.

omod/src/test/java/org/openmrs/module/authentication/web/EmailAuthenticationSchemeTest.java
mseaton: Fixed

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
mogoodrich: Makes sense, good call

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
mogoodrich: My personal preference is in this subject as well... but, if it's bad practice, I can live with it. :)

omod/src/main/java/org/openmrs/module/authentication/web/EmailAuthenticationScheme.java
mogoodrich: This makes sense to me.

mogoodrich (commented):
Thanks @mseaton 

And thanks @ibacher  for the detailed review, very helpful.

Side note, I'm generally in favor of more eyes and less privacy, so I'm not necessarily advocating for this, but reading all these comments made me think: should this repo be private/restricted in any way?

PR #52, opened 2026-06-11
AUT-26: Return 401 with Location header for unauthenticated and partially authenticated users on REST endpoints

dilankavishka:
## Requirements

- [x] My work includes tests or is validated by existing tests.

## Summary

Previously, the authentication filter would return HTTP 200 from the session endpoint even for unauthenticated users. This PR updates the behavior so that when an unauthenticated or partially authenticated user hits /ws/rest/v1/session, the filter returns 401 Unauthorized with a Location header pointing to the configured challenge URL (login or 2FA setup screen).

- Returns a 401 with a Location header pointing to the configured challenge URL from the session endpoint (/ws/rest/v1/session)
- Returns a 401 instead of a 302 redirect for all other REST endpoints

## Screenshots
Screenshot 2026-06-11 at 17 54 23

## Related Issue

https://openmrs.atlassian.net/browse/AUT-26

## Other
N/A

### DIFF
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -130,7 +130,6 @@ public class AuthenticationFilter implements Filter {
 					// If any credentials were passed in the request or session attempt to authentication with them
 					AuthenticationCredentials credentials = webScheme.getCredentials(session);
 					String challengeUrl = WebUtil.contextualizeUrl(request, webScheme.getChallengeUrl(session));
-					String o3ChallengeUrl = WebUtil.contextualizeUrl(request, webScheme.getO3ChallengeUrl(session));
 					if (credentials != null) {
 						try {
 							session.removeErrorMessage();
@@ -150,10 +149,7 @@ public class AuthenticationFilter implements Filter {
 					}
 					// If no credentials were found, redirect to challenge url unless whitelisted
 					else {
-						if (WebUtil.isO3SpaRequest(request)) {
-							handleAuthenticationFailure(request, response, o3ChallengeUrl);
-						}
-						else if (!WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getWhiteList())) {
+						if (!WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getWhiteList())) {
 							log.trace("Authentication required: " + request.getRequestURI());
 							handleAuthenticationFailure(request, response, challengeUrl);
 						}
@@ -176,7 +172,7 @@ public class AuthenticationFilter implements Filter {
 	 * @param challengeUrl the challengeUrl to direct the response to
 	 */
 	protected void handleAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, String challengeUrl) throws IOException {
-		if (WebUtil.isO3SpaRequest(request) || WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getNonRedirectUrls())) {
+		if (WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getNonRedirectUrls())) {
 			response.setHeader("Location", challengeUrl);
 			response.sendError(HttpServletResponse.SC_UNAUTHORIZED);
 		}
--- a/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
@@ -149,10 +149,7 @@ public class AuthenticationFilter implements Filter {
 					}
 					// If no credentials were found, redirect to challenge url unless whitelisted
 					else {
-						if (WebUtil.isRestSessionEndpointRequest(request)) {
-							handleAuthenticationFailure(request, response, challengeUrl);
-						}
-						else if (!WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getWhiteList())) {
+						if (!WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getWhiteList())) {
 							log.trace("Authentication required: " + request.getRequestURI());
 							handleAuthenticationFailure(request, response, challengeUrl);
 						}
@@ -175,7 +172,7 @@ public class AuthenticationFilter implements Filter {
 	 * @param challengeUrl the challengeUrl to direct the response to
 	 */
 	protected void handleAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, String challengeUrl) throws IOException {
-		if (WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getNonRedirectUrls())) {
+		if (WebUtil.isRestSessionEndpointRequest(request) || WebUtil.urlMatchesAnyPattern(request, AuthenticationConfig.getNonRedirectUrls())) {
 			response.setHeader("Location", challengeUrl);
 			response.sendError(HttpServletResponse.SC_UNAUTHORIZED);
 		}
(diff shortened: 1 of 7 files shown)


dilankavishka:
@jayasanka-sack @janithRS Could you review this PR please?

omod/src/main/java/org/openmrs/module/authentication/web/WebAuthenticationScheme.java
```diff
@@ -121,6 +121,16 @@ public Authenticated authenticate(Credentials credentials) throws ContextAuthent
      */
     public abstract String getChallengeUrl(AuthenticationSession session);
 
+    /**
+     * Returns the challenge url sent in the Location header of a 401 response to O3 clients.
+     * Defaults to getChallengeUrl. Override (e.g. in BasicWithLocationAuthenticationScheme) to
+     * return a dedicated O3 destination such as a TOTP setup screen.
+     * @param session the current AuthenticationSession
+     */
+    public String getO3ChallengeUrl(AuthenticationSession session) {
```
ibacher: Why do we need a separate "O3ChallengeUrl"? It seems like we should be able to reuse the existing `challengeUrl` concept for this.

omod/src/main/java/org/openmrs/module/authentication/web/WebUtil.java
```diff
@@ -64,4 +64,14 @@ public static String contextualizeUrl(HttpServletRequest request, String url) {
...
         return url;
     }
+    
+    /**
+     * Checks if the request is coming from the O3 Frontend (REST session endpoint)
+     *
+     * @param request the HttpServletRequest to check
+     * @return true if the request is from O3 and should receive a 401 rather than a redirect
+     */
+    public static boolean isO3SpaRequest(HttpServletRequest request) {
```
ibacher: I also think this is poorly named. This is really just checking if it's hitting the session endpoint which any REST client might end-up doing, not just O3.

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
```diff
@@ -149,9 +150,8 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 					}
 					// If no credentials were found, redirect to challenge url unless whitelisted
 					else {
-						if (WebUtil.matchesPath(request, "/ws/rest/*/session")) {
-							// Add a location header to the session endpoint to support frontend redirection to login
-							response.setHeader("Location", challengeUrl);
+						if (WebUtil.isO3SpaRequest(request)) {
```
ibacher: I'd argue that the swap here makes it harder to reason about how this works. The `matchesPath()` call makes this code easier to read.

ibacher: Of course, if we're just routing to the same place, is there any reason we need this `if`?

omod/src/main/java/org/openmrs/module/authentication/web/WebUtil.java
janithRS: Are we planning on doing anything specific to O3 here? (because I think it was mentioned that the REST API will be passed through this, whether or not it's O3). If not, I think we could update the naming as @ibacher mentioned.

omod/src/main/java/org/openmrs/module/authentication/web/AuthenticationFilter.java
dilankavishka: Hey @ibacher!

I have a question. If I understood you correctly, are you suggesting we should remove `/ws/rest/v1/session` from the default whitelist in [`AuthenticationConfig.getWhiteList()`](https://github.com/openmrs/openmrs-module-authentication/blob/58592bba56f037262a0dcfe633a93831a4a4fa21/api/src/main/java/org/openmrs/module/authentication/AuthenticationConfig.java#L253) and drop the special **if** block entirely from doFilter?
Is that what you had in mind?

cc: @jayasanka-sack @janithRS

ibacher: Yes. With these changes the `if` and `else if` are functionally equivalent so if we can reduce them, we're making things more maintainable.

dilankavishka: Understood!

omod/src/main/java/org/openmrs/module/authentication/web/WebAuthenticationScheme.java
dilankavishka: Removed this to use the existing challengeUrl (`getChallengeUrl`)

omod/src/main/java/org/openmrs/module/authentication/web/WebUtil.java
dilankavishka: Renamed to `isRestSessionEndpointRequest`

dilankavishka: I got the point, we need to check if any client hitting the `/ws/rest/v1/session` (O3 or otherwise) should gets the same 401. If i'm wrong, please do correct me. 
I updated the name to `isRestSessionEndpointRequest`

janithRS (approved):
LGTM 👍🏽 
BTW, avoid adding commit messages like the last two here in the future. Let's go with more descriptive ones :)

jayasanka-sack:
Thanks for the PR @dilankavishka !  ❤️

I'm just trying to understand the approach here. The objective of this work as I understand it, is to:

when the user is partially logged in:
1. Return a 401 (with the appropriate Location header) from the session endpoint 
2. Return a 401 on REST endpoints instead of a 302 redirect.

Looking through the changes I don't think this PR addresses those requirements. Could you help me understand how these changes relate to the intended objective?

thanks!

omod/src/main/java/org/openmrs/module/authentication/web/WebUtil.java
janithRS: I have 2 questions
1. Where are we using this util?
2. This util suggests that it returns true if the request path is /ws/rest/v1/session, but the util name is isRestSessionEndpointRequest. Does that condition qualify the function body?

mseaton (changes requested):
I'd like to better understand the nature of this work.  It seems to me like this is actually undoing some of what was done to actually solve the issue that this describes that it is solving over a year ago?  See the work done here:  https://openmrs.atlassian.net/browse/AUT-9 .  I need to better understand what is being proposed here and why, and why we think the current code as-is doesn't function as needed before this can be approved.

dilankavishka:
> I'd like to better understand the nature of this work. It seems to me like this is actually undoing some of what was done to actually solve the issue that this describes that it is solving over a year ago? See the work done here: https://openmrs.atlassian.net/browse/AUT-9 . I need to better understand what is being proposed here and why, and why we think the current code as-is doesn't function as needed before this can be approved.

Hi @mseaton, thank you for the comment on this PR.

I had a discussion with my mentors today and we've come to the conclusion that the backend changes here are not the right approach. We've decided not to make any changes to the backend for now. Instead, the frontend will handle this like, when the session endpoint is called and the response includes a `Location` header, the frontend will redirect to that location. That is our current goal.

I also have a question about configuring TOTP and I'll post about it on the talk thread in more detail and mention you there and I would appreciate your help on that as well.

@jayasanka-sack shall I close this PR as we've changed the approach?

I learned a lot through this process though, so I believe it wasn't wasted effort!
Thank you so much @ibacher @jayasanka-sack @janithRS

ibacher:
PRs are never wasted effort as long as we're learning from them. I have many more abandoned spikes on things in OMRS than I've ever actually committed, for whatever that's worth.

dkayiwa:
The "99 ways" concept stems from a famous Thomas Edison quote about his relentless trial and error in pursuit of the incandescent light bulb. Rather than failing, he treated every misstep as discovering a way not to make 😊

janithRS:
Good effort and learning opportunity indeed @dilankavishka ! ( for us too ;) Now let's focus on the new approach.

PR #53, opened 2026-06-29
AUT-27: Implement new endpoints for 2FA enrollment

dilankavishka:
## Requirements

- [ ] My work includes tests or is validated by existing tests.

## Summary

I'm raising this PR to Introduce a REST API controller for 2FA (TOTP) enrollment and verification.

1. Endpoints Added:

`POST /ws/rest/v1/auth/{schemeId}/enrollment` (Initiate)
- generates a new secret key for the logged in (authenticated) user.
- keeps the secret key temporarily in the session.
- Returns the secret and a QR code URI.

`POST /ws/rest/v1/auth/{schemeId}/enrollment/verify `(Verify)
- accepts the code (6-digit verification code).
- verifies the code against the stored session secret.

2. Dependencies:

Added web services REST dependencies to `pom.xml` and `omod/pom.xml`

## Screenshots
N/A

## Related Issue

https://openmrs.atlassian.net/browse/AUT-27

## Other
N/A


dilankavishka:
@jayasanka-sack @ibacher  @mseaton @janithrs :- I'm opening this PR as a draft since I'm still working on this. In the meantime, I would like to hear any suggestions or feedback on the work upto now. so far, I've tested the enrollment endpoints with Postman, and they're working as expected.
1. when I call the `enrollment` endpoint it return the secret and the qrcodeuri
2. I decoded that base64 and scan the QR through my authenticator app
3. I send that code as a query param with `enrollment/verify` endpoint and It gives me the isVerifyCode as true.

Moving forward with the `enrollment/verify` endpoint, I have to encrypt the secret and save it as a user property permanently.

Screenshot 2026-06-29 at 21 51 50
Screenshot 2026-06-29 at 21 51 43

jayasanka-sack:
Thanks for the PR @dilankavishka! 

I think the controller is too TOTP-specific. The controller should not know about 	`TotpAuthenticationScheme`	. Since the URL already has 	`{schemeId}`	, it should resolve the matching `WebAuthenticationScheme` and delegate enrollment to that scheme.

so something like: `scheme.initiateEnrollment(....);`

You can have it in WebAuthenticationScheme with a default "enrollment not supported" implementation, then override it only in TotpAuthenticationScheme for this PR. You can implement it in other relevant schemes in future PRs.

That way the API stays generic and future authentication schemes can add their own enrollment flow without changing the controller.

dilankavishka:
Thanks for reviewing my PR @jayasanka-sack!

I've updated the code with understanding your comment and approach and if i'm correct, the approach is open closed design principle (part of the solid design) right?

And i've checked with configuring another scheme which is basic2 and with the error handling now if we send basic2 as the scheme it will gives the proper status code with the error message:
Screenshot 2026-07-01 at 20 41 49

could you please take a look at this again?

cc: @ibacher @mseaton @janithrs

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
```diff
@@ -0,0 +1,57 @@
...
+	}
+	
+	@ExceptionHandler(UnsupportedOperationException.class)
+	@ResponseStatus(HttpStatus.METHOD_NOT_ALLOWED)
+	@ResponseBody
+	public SimpleObject handleUnsupportedException(UnsupportedOperationException exception) {
+		SimpleObject error = new SimpleObject();
+		error.put("message", exception.getMessage());
+		return error;
+	}
```
dkayiwa: This needs fixing before merge: every `IllegalArgumentException` these endpoints raise comes back to the client as HTTP 500. The common one is a mistyped code. `verifyEnrollment` does `throw new IllegalArgumentException("Invalid code Entered")`, and since this controller maps only `UnsupportedOperationException`, that falls through to `BaseRestController`'s catch-all handler, which sets `SC_INTERNAL_SERVER_ERROR` and logs it at error level. The "no pending secret" and "unsupported scheme type" paths return 500 the same way. A sibling handler maps them to 400:

```suggestion
	@ExceptionHandler(UnsupportedOperationException.class)
	@ResponseStatus(HttpStatus.METHOD_NOT_ALLOWED)
	@ResponseBody
	public SimpleObject handleUnsupportedException(UnsupportedOperationException exception) {
		SimpleObject error = new SimpleObject();
		error.put("message", exception.getMessage());
		return error;
	}

	@ExceptionHandler(IllegalArgumentException.class)
	@ResponseStatus(HttpStatus.BAD_REQUEST)
	@ResponseBody
	public SimpleObject handleIllegalArgumentException(IllegalArgumentException exception) {
		SimpleObject error = new SimpleObject();
		error.put("message", exception.getMessage());
		return error;
	}
```

Within this controller the only `IllegalArgumentException`s are the ones the code throws deliberately, so 400 fits all of them.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +219,51 @@ public String getClientName() {
 			return user == null ? null : user.getUsername();
 		}
 	}
+	
+	/**
+	 * Generates a new TOTP secret and a QR code URI, then stashes the secret in the HTTP session for verification.
+	 */
+	@Override
+	public SimpleObject initiateEnrollment(HttpServletRequest request) {
```
dkayiwa: This one also needs handling before merge: `Context.getAuthenticatedUser()` can return null, and `user.getUsername()` on the next line then throws an NPE that surfaces as HTTP 500. It is reachable: the REST `AuthorizationFilter` explicitly "never stops execution" and calls `chain.doFilter` even when no one is authenticated, so an anonymous `POST .../enrollment` gets here with a null user. That should be a 401, not a 500. Guarding and throwing `APIAuthenticationException` (which `BaseRestController` turns into 401 when not logged in) does it:

```java
if (user == null) {
	throw new APIAuthenticationException("Must be authenticated to enroll");
}
```

(needs `import org.openmrs.api.APIAuthenticationException;`). `verifyEnrollment` has the same null-user gap, though there it is partly shielded because `setUserProperty` runs its own privilege check.

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * REST Controller to manage Two-Factor Authentication enrollment and verification.
+ */
+@Controller
+@RequestMapping(value = "/rest/v1/auth/{schemeId}")
+public class TwoFactorEnrollmentController extends BaseRestController {
```
dkayiwa, line 44: These endpoints ship without any tests (the PR checklist flags it, and SonarCloud shows 0% coverage on the new code). For an auth feature I would want that covered before merge, and `TotpAuthenticationSchemeTest` (it extends `BaseWebAuthenticationTest` and already drives this scheme with a mock request) is a natural home. Worth asserting: initiate returns a secret and QR URI and stashes the secret in the session; verify with a code generated for that secret saves the encrypted user property and clears the session; and the failure paths (no pending secret, wrong code, a scheme that does not override enrollment) return what you expect. That last group would have caught the 500 status issues flagged above.

omod/src/main/java/org/openmrs/module/authentication/web/WebAuthenticationScheme.java
```diff
@@ -18,7 +18,10 @@
 import org.openmrs.module.authentication.ConfigurableAuthenticationScheme;
 import org.openmrs.module.authentication.UserLogin;
 import org.openmrs.module.authentication.UserLoginTracker;
+import org.openmrs.module.webservices.rest.SimpleObject;
+import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOperationException;
```
dkayiwa: This import isn't used. Both new methods throw `UnsupportedOperationException`, not `ResourceDoesNotSupportOperationException`, so it can come out (likely the single issue SonarCloud flagged).

omod/pom.xml
```diff
@@ -71,7 +71,67 @@
...
+            <version>${webservices.restVersion}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.openmrs.module</groupId>
+            <artifactId>webservices.rest-omod-2.2</artifactId>
+            <version>${webservices.restVersion}</version>
+            <scope>provided</scope>
+        </dependency>
```
dkayiwa: I don't think these eight version-specific modules are needed. The only REST types this PR compiles against (`SimpleObject` and `BaseRestController`, plus the `ResourceDoesNotSupportOperationException` import that is currently unused) all live in `webservices.rest-omod-common`, which you already declare just above. So `webservices.rest-omod-1.8` through `-2.2` look droppable, leaving only `omod-common`. Worth a quick build to confirm, but trimming them keeps the dependency tree tidy. Non-blocking.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +219,51 @@ public String getClientName() {
...
+		User user = Context.getAuthenticatedUser();
+		String secret = generateSecret();
```
dkayiwa, line 246: The pending secret is stashed in the HTTP session here and read back in `verifyEnrollment`, so both calls have to hit the same session. That holds for a browser or Postman, where the `JSESSIONID` cookie is carried automatically, but a REST client that authenticates per request without keeping that cookie would land on a fresh session at verify time and get "No pending enrollment totp secret found". Is the intended consumer (the O3 frontend?) cookie-based? If it needs to work for stateless clients too, returning the secret to the caller and having them send it back on verify would remove the session dependency. Not blocking, just want to make sure the clients you care about are covered.

omod/pom.xml
dilankavishka: Updated with keeping only the `omod-common` and confirmed with a build.

omod/src/main/java/org/openmrs/module/authentication/web/WebAuthenticationScheme.java
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: applied null guard for both enrollment and verifyEnrollment:

Screenshot 2026-07-02 at 06 08 19

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dilankavishka: Added new handler method and now it comes as 400 bad request.
Screenshot 2026-07-02 at 06 17 51

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
```diff
@@ -0,0 +1,66 @@
...
+	public SimpleObject initiateEnrollment(@PathVariable("schemeId") String schemeId, HttpServletRequest request) {
+		WebAuthenticationScheme authScheme = getWebAuthenticationScheme(schemeId);
+		return authScheme.initiateEnrollment(request);
+	}
+	
+	@RequestMapping(method = RequestMethod.POST, value = "/enrollment/verify")
+	@ResponseBody
+	public SimpleObject verifyEnrollment(@PathVariable("schemeId") String schemeId, @RequestBody SimpleObject payload, HttpServletRequest request) {
+		WebAuthenticationScheme authScheme = getWebAuthenticationScheme(schemeId);
+		return authScheme.verifyEnrollment(payload, request);
```
jayasanka-sack: I think catching the exception here instead of adding an Exception handler would be better as we don't want to catch whatever UnsupportedExceptions thrown by any method but to specifically thrown by `verifyEnrollment` method. What do you think?

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
jayasanka-sack: I'd suggest adding tests rather than relying on Postman. Nothing wrong with using Postman to verify things, but having to redeploy every time you make a change gets old pretty quickly.

Tests will make development much faster, and we generally encourage TDD since it also helps catch regressions later. Also, no need to post screenshots. A passing test is much better evidence that it works.

dilankavishka: > I'd suggest adding tests rather than relying on Postman. Nothing wrong with using Postman to verify things, but having to redeploy every time you make a change gets old pretty quickly.
> 
> Tests will make development much faster, and we generally encourage TDD since it also helps catch regressions later. Also, no need to post screenshots. A passing test is much better evidence that it works.

I'm on it jaye! I will update the PR ASAP with the test cases :)

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dilankavishka: @jayasanka-sack, If i understood correctly, you're suggesting that to use a try-catch block inside the enrollment and enrollement/verify and handle that instead of the `@ExceptionHandler(UnsupportedOperationException.class)`. Am I correct?

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +220,61 @@ public String getClientName() {
...
+	
+	/**
+	 * Generates a new TOTP secret and a QR code URI, then stashes the secret in the HTTP session for verification.
+	 */
+	@Override
+	public SimpleObject initiateEnrollment(HttpServletRequest request) {
+		User user = Context.getAuthenticatedUser();
+		
+		if (user == null) {
+			throw new APIAuthenticationException("Must be authenticated user to enroll 2FA");
```
janithRS: I remember we discussed about using intl for messages so that we can easily support multiple langs and locales if needed. Was it this place?

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: @dkayiwa Yes! our primary intended consumer for this enrollment flow is the O3 frontend.

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dilankavishka: Updated with using the try-catch

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dilankavishka: Wrote the unit tests and updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +220,65 @@ public String getClientName() {
...
+		
+		String temporarySavedSecret = (String) request.getSession().getAttribute(PENDING_ENROLLMENT_SECRET);
+		
+		if (temporarySavedSecret == null) {
+			throw new IllegalArgumentException("No pending enrollment totp secret found");
+		}
+		
+		String code = payload.get("code");
+		
+		boolean isValidCode = verifyCode(temporarySavedSecret, code);
```
dkayiwa: One more 500 to squash before this merges: a verify request with a missing `code` still comes back as an internal server error. `payload.get("code")` returns null when the field is absent, and `DefaultCodeVerifier.timeSafeStringComparison` NPEs on `code.getBytes()` (its try block only catches `CodeGenerationException`), so the request lands in the `BaseRestController` catch-all as a 500 with an error-level stack trace in the log. Same outcome if a client sends the code as a JSON number, `{"code": 123456}`, which a 6-digit code invites: Jackson hands `SimpleObject` an Integer and the implicit cast to String throws `ClassCastException`. Both are client mistakes, so they should get the same 400 treatment the other bad inputs now get:

```suggestion
		Object rawCode = payload.get("code");
		if (rawCode == null || StringUtils.isBlank(rawCode.toString())) {
			throw new IllegalArgumentException("A verification code is required");
		}
		
		boolean isValidCode = verifyCode(temporarySavedSecret, rawCode.toString());
```

The `toString()` also lets the numeric case verify correctly instead of blowing up. Worth a companion test asserting that a payload without a `code` gets an `IllegalArgumentException`.

omod/pom.xml
```diff
@@ -71,7 +71,12 @@
             <version>3.1.0</version>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.openmrs.module</groupId>
+            <artifactId>webservices.rest-omod-common</artifactId>
+            <version>${webservices.restVersion}</version>
+            <scope>provided</scope>
+        </dependency>
```
dkayiwa, line 79: Before this merges, config.xml should declare what is now a hard runtime dependency. This is the module's first compile dependency on webservices.rest classes, and config.xml only lists that module under `aware_of_modules` (added for filter ordering), which orders startup but does not require the module to be present. On a server that doesn't run webservices.rest, the component scan of `org.openmrs.module.authentication.web` still instantiates `TwoFactorEnrollmentController` at context refresh, resolving its `BaseRestController` superclass throws `NoClassDefFoundError`, and the whole authentication module fails to start. Logins then quietly fall back to core's username/password scheme, meaning 2FA is off and nothing but a startup log line says so. A REST-less server is admittedly rare, but the insurance is one line in config.xml:

```xml
<require_module>org.openmrs.module.webservices.rest</require_module>
```

With that, the module manager reports a clear missing-required-module message instead of a cryptic startup failure. The alternative, keeping the module usable without REST by loading the controller from a conditional resource jar, is more machinery than this needs. I'd declare the requirement.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +220,65 @@ public String getClientName() {
...
+		if (!isValidCode) {
+			throw new IllegalArgumentException("Invalid code Entered");
+		}
+		
+		saveSecret(user, temporarySavedSecret);
```
dkayiwa, line 300: Was leaving activation out of this flow deliberate? After a successful verify the secret is saved, but nothing sets the `authentication.secondaryType` user property, and that property is the only thing `TwoFactorAuthenticationScheme` consults when deciding whether to demand a second factor. So a user who completes enrollment here still logs in with just a password until some separate call sets that property. If the plan is for the O3 frontend to set it on its own, there's an ordering footgun to be aware of: set the property before a completed verify (or abandon enrollment after setting it) and the user is locked out at the next login, since the secondary challenge is unconditional and `authenticate` throws `authentication.error.noSecretConfiguredForUser` when no secret exists (`isUserConfigurationRequired` is never consulted by the login flow). My preference would be for verify to also append this scheme's id to `authentication.secondaryType` in the same save, so enrollment and activation are atomic and clients can't get the order wrong. If activation is meant to stay a separate concern for AUT-27, fine, but let's make that a deliberate call.

omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
```diff
@@ -114,4 +116,124 @@ public void shouldUseConfiguredHeaderName() {
 		assertThat(credentials, notNullValue());
 		assertThat(credentials.getClientName(), equalTo("testing"));
 	}
+	
+	@Test
+	public void shouldInitiateEnrollmentSuccessfully() {
+		org.openmrs.api.context.Context.setUserContext(new MockUserContext(candidateUser));
```
dkayiwa: Tiny style thing, take or leave: the new tests spell out fully qualified names inline (`org.openmrs.api.context.Context`, the controller, `ResourceDoesNotSupportOperationException`) where the rest of the file uses imports. Plain imports would read better.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Of course @dkayiwa, that was in my mind as well. shall we keep the scope of this PR strictly to the enrollment APIs? and should we address the activation in a separate ticket? what do you think or would you prefer I add it to this PR?

omod/pom.xml
dilankavishka: Update with wrapping this inside the `<require_modules>`

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Handled this and add unit tests as well

omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
dilankavishka: I took this :) and cleaned the imports

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Thank you for pointing this out.
I have replaced the hardcoded error strings with the translation keys and registered them in `messages.properties`

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +220,68 @@ public String getClientName() {
 			return user == null ? null : user.getUsername();
 		}
 	}
+	
+	/**
+	 * Generates a new TOTP secret and a QR code URI, then stashes the secret in the HTTP session for verification.
+	 */
+	@Override
+	public SimpleObject initiateEnrollment(HttpServletRequest request) {
```
dkayiwa: Was it deliberate that enrollment still works when the user already has a secret configured? As written, any request riding an authenticated session can call initiate and then verify, and silently replace the stored TOTP secret. For a user whose second factor is already active, that means a hijacked session (or XSS in the SPA) can move 2FA to the attacker's device without ever presenting the current code, and the real user's authenticator quietly stops working at their next login. Since `isUserConfigurationRequired(user)` already detects an existing secret, I'd refuse enrollment in that case, and leave secret rotation to a future flow that first proves control of the current factor (or to an admin reset). If first-time enrollment is the intended scope for now, that guard is a couple of lines.

api/src/main/resources/messages.properties
```diff
@@ -18,4 +18,8 @@ authentication.error.locationRequired=You must choose a location
 authentication.error.invalidLocation=Not allowed to log in at location {0}
 authentication.error.invalidPassword=Incorrect Password
 authentication.email.subject=Your verification code
-authentication.email.body=Your verification code is {0}
\ No newline at end of file
+authentication.email.body=Your verification code is {0}
+authentication.error.mustBeAuthenticatedToEnroll=Must Be Authenticated To Enroll
+authentication.error.noPendingEnrollmentTotpSecretFound=No Pending Enrollment TOTP Found
+authentication.error.requiredVerificationCode=Required A Verification Code
+authentication.error.invalidCodeEntered=Invalid Code Entered
```
dkayiwa: Take these or leave them, but a few of the new messages read awkwardly in English, "Required A Verification Code" especially:

```suggestion
authentication.error.mustBeAuthenticatedToEnroll=You must be authenticated to enroll
authentication.error.noPendingEnrollmentTotpSecretFound=No pending enrollment found
authentication.error.requiredVerificationCode=A verification code is required
authentication.error.invalidCodeEntered=Invalid verification code
```

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dkayiwa: I'd add it to this PR rather than a separate ticket. The building block is already there: `TwoFactorAuthenticationScheme.addSecondaryAuthenticationSchemeForUser(user, schemeId)` does the append logic, so `verifyEnrollment` can set that property in the same save as the secret, with one test asserting it. My main reason is that until activation is wired somewhere, a user who scans the QR and gets back `isValidCode` will reasonably believe 2FA now protects their account, when nothing about their login has actually changed. That silent gap worries me more than the small increase in scope. Un-enrollment and admin reset can stay a separate ticket.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
jayasanka-sack: Good point on the session dependency Daniel. The issues is.. sending the secret back from the client during verify defeats the trust boundary. Because the client could tamper with that value and effectively control what the backend verifies against.

A stateless version would probably need a signed short-lived token/JWT or similar, where the backend can verify the enrollment state was issued by the server. But we discussed this on the Platform team call with Mike and others, and the decision was to rely on the server session for this flow. Since this flow isn't intended to support non-cookie-based clients anyway, storing the temporary secret in the session is the preferred approach here.

One concern I do have with the session approach is that the pending enrollment remains valid until the session expires, which could be much longer than necessary. @dilankavishka it might be worth storing a second session attribute with the enrollment initiation timestamp and rejecting verification if it's older than, say, 2 minutes. That keeps the enrollment window short while still relying on the session.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +220,68 @@ public String getClientName() {
...
+		}
+		
+		Object rawCode = payload.get("code");
+		if (rawCode == null || StringUtils.isBlank(rawCode.toString())) {
+			throw new IllegalArgumentException("authentication.error.requiredVerificationCode");
+		}
+		
+		boolean isValidCode = verifyCode(temporarySavedSecret, rawCode.toString());
+		if (!isValidCode) {
+			throw new IllegalArgumentException("authentication.error.invalidCodeEntered");
```
jayasanka-sack: IllegalArgumentException might not be the right call here (and above). It is generally used when a method is called incorrectly by its caller (the programmer passed an invalid argument). Here, the method is handling user input from an HTTP request, so invalid input is an expected runtime condition rather than a programming error.

dilankavishka: Updated and used `IllegalRequestException`

dilankavishka: Updated with using `IllegalRequestException`

api/src/main/resources/messages.properties
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: updated @jayasanka-sack and I used concurrent time units for the duration

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +225,93 @@ public String getClientName() {
...
+	protected void saveSecret(User user, String secret) {
+		String encryptedSecret = Security.encrypt(secret);
+		Context.getUserService().setUserProperty(user, getSecretUserPropertyName(), encryptedSecret);
+		
+		AuthenticationScheme twoFactor = AuthenticationConfig.getAuthenticationScheme("twofactor");
+		if (twoFactor instanceof TwoFactorAuthenticationScheme) {
+			((TwoFactorAuthenticationScheme) twoFactor).addSecondaryAuthenticationSchemeForUser(user, getSchemeId());
+			Context.getUserService().saveUser(user);
+		}
+	}
```
dkayiwa, line 315: This needs fixing before it can merge. `saveSecret` calls `Context.getUserService().saveUser(user)` directly, and `saveUser` requires the "Edit Users" privilege (`UserServiceImpl.saveUser` calls `Context.requirePrivilege(EDIT_USERS)`). A normal user enrolling their own second factor doesn't hold that privilege, so verify fails right here with a 403, after the secret was already persisted on the `setUserProperty` line just above.

If merged as-is, a non-admin who enrolls hits this sequence: the secret is saved (that line works because core wraps its internal save in a temporary `EDIT_USERS` proxy privilege when you edit your own account), then `saveUser` throws `APIAuthenticationException`, the client gets a 403, and because the secret is now stored `isUserConfigurationRequired` returns false, so retrying `initiate` is refused with `secretAlreadyConfigured`. The user ends up with a stored secret, no active second factor, and no way back through this flow.

The suite doesn't catch it because `MockTotpAuthenticationScheme` overrides `saveSecret` with an in-memory `user.setUserProperty(...)` and no service call, so `shouldVerifyEnrollmentSuccessfully` never runs this code. A test that drives the real `saveSecret` as a non-privileged user would fail today.

Persisting through `setUserProperty` (which self-elevates) instead of a bare `saveUser` fixes it. Setting the secondary type before the secret save lets the single `setUserProperty` call persist both:

```suggestion
	protected void saveSecret(User user, String secret) {
		String encryptedSecret = Security.encrypt(secret);

		AuthenticationScheme twoFactor = AuthenticationConfig.getAuthenticationScheme("twofactor");
		if (twoFactor instanceof TwoFactorAuthenticationScheme) {
			((TwoFactorAuthenticationScheme) twoFactor).addSecondaryAuthenticationSchemeForUser(user, getSchemeId());
		}

		// setUserProperty grants a temporary EDIT_USERS proxy privilege and saves the whole user, so this one
		// call persists both the secret and the secondary type set above. A bare saveUser(user) would instead
		// require the enrolling user to hold EDIT_USERS, which ordinary users don't have.
		Context.getUserService().setUserProperty(user, getSecretUserPropertyName(), encryptedSecret);
	}
```

Separately, and worth sorting out in this PR: the two-factor scheme id is hardcoded as `"twofactor"`, but the README registers that scheme under other ids (its examples use `2fa` and `secret`). `getAuthenticationScheme("twofactor")` returns null when nothing is registered under exactly that id, so `instanceof` is false and activation is silently skipped: the secret is saved but `authentication.secondaryType` is never set, and the user logs in with only a password while believing 2FA is on. On any deployment that doesn't name its two-factor scheme `twofactor`, this won't activate. Resolving the active scheme (`AuthenticationConfig.getAuthenticationScheme()`) rather than a fixed id would be robust (the existing test would then configure the two-factor scheme as the active one instead of `totp`). At a minimum, please confirm the target config uses `twofactor`.

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+	
+	@ExceptionHandler(IllegalRequestException.class)
+	@ResponseStatus(HttpStatus.BAD_REQUEST)
+	@ResponseBody
+	public SimpleObject handleIllegalRequestException(IllegalRequestException exception) {
+		SimpleObject error = new SimpleObject();
+		error.put("message", exception.getMessage());
+		return error;
+	}
```
dkayiwa: Non-blocking, but I'd remove this handler rather than keep it. It returns `{"message": exception.getMessage()}`, which is worse than letting the framework handle the exception, in two ways:

- The message is the raw key. `IllegalRequestException("authentication.error.invalidCodeEntered")` reaches the client as that literal string. Routed through `BaseRestController` instead, the exception goes to `RestUtil.wrapErrorResponse`, which resolves the key against `messages.properties` and returns it as `translatedMessage`. That resolution is what actually delivers the localization the earlier i18n thread was after; this handler bypasses it.
- `{"message": ...}` isn't the standard REST error envelope (`{"error": {"message", "rawMessage", "translatedMessage", "code", "detail"}}`) that the rest of the API returns, so a client parsing the usual shape won't find the message where it expects it.

`IllegalRequestException` is already annotated `@ResponseStatus(BAD_REQUEST)`, so deleting this handler makes the framework return a 400 with the standard, localized envelope, which covers both points at once.

dilankavishka: Removed.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Updated.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +225,93 @@ public String getClientName() {
...
+		}
+		
+		String temporarySavedSecret = (String) request.getSession().getAttribute(PENDING_ENROLLMENT_SECRET);
+		
+		if (temporarySavedSecret == null) {
+			throw new IllegalRequestException("authentication.error.noPendingEnrollmentTotpSecretFound");
+		}
+		
+		Long initiationTime = (Long) request.getSession().getAttribute(PENDING_ENROLLMENT_TIME);
+		long maxLifetimeForSecret = TimeUnit.MINUTES.toMillis(2);
```
dkayiwa: Now that every other knob on this scheme (secretLength, codeValidityPeriod, allowedDiscrepancy and friends) comes through configure(), consider making this window a config property too, something like an enrollmentWindowSeconds read with AuthenticationUtil.getInteger and defaulting to 120 to keep the behavior you and Jayasanka agreed. Two minutes is fine for someone who already has an authenticator app, but a first time user who needs to install one first can run over it, and the only recovery is bouncing through initiate again for a fresh QR to rescan. With a property, deployments that onboard novice users can widen it without a code change. Happy for this to be a follow-up if you'd rather freeze this PR.

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
```diff
@@ -0,0 +1,58 @@
...
+import org.openmrs.module.webservices.rest.web.v1_0.controller.BaseRestController;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
```
dkayiwa: Leftover from the removed handler: HttpStatus, ExceptionHandler and ResponseStatus are no longer used.

```suggestion
import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.PathVariable;
import org.springframework.web.bind.annotation.RequestBody;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RequestMethod;
import org.springframework.web.bind.annotation.ResponseBody;
```

omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
```diff
@@ -114,4 +124,181 @@ public void shouldUseConfiguredHeaderName() {
...
+		
+		candidateUser.setUserProperty(authenticationScheme.getSecretUserPropertyName(), "EXISTING_SECRET");
+		
+		assertThrows(IllegalRequestException.class, () -> {
+			authenticationScheme.initiateEnrollment(request);
+		});
+	}
+	
+	@Test
+	public void ShouldThrowExceptionIfSessionIsExpired() {
```
dkayiwa: `ShouldThrowExceptionIfSessionIsExpired` starts with a capital S, and `shouldVerifyEnrollmentSuccessfulWithNumericCode ()` has a stray space before the parentheses.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dkayiwa: The mock still overrides saveSecret wholesale, so the production method now has zero test execution. I checked by putting a throw at the top of the real saveSecret and running the suite: all 126 omod tests still pass. That means the activation assertion in shouldVerifyEnrollmentSuccessfully is exercising the mock's duplicated copy of this logic, and a future edit that breaks the real method (like the saveUser call this thread started with) would again go unnoticed. I'd shrink the override surface: keep activation in the real method and extract just the service call into a hook the mock overrides.

```java
protected void saveSecret(User user, String secret) {
	String encryptedSecret = Security.encrypt(secret);
	
	AuthenticationScheme twoFactor = AuthenticationConfig.getAuthenticationScheme();
	if (twoFactor instanceof TwoFactorAuthenticationScheme) {
		((TwoFactorAuthenticationScheme) twoFactor).addSecondaryAuthenticationSchemeForUser(user, getSchemeId());
	}
	
	persistSecret(user, encryptedSecret);
}

protected void persistSecret(User user, String encryptedSecret) {
	Context.getUserService().setUserProperty(user, getSecretUserPropertyName(), encryptedSecret);
}
```

MockTotpAuthenticationScheme then overrides only persistSecret with the in-memory user.setUserProperty, and the scheme lookup, the instanceof check and the ordering all run for real under test. Everything in that path already works in this test setup, since the mock's current copy calls the same methods. Not blocking, but it's a small change and this spot has already been bitten once by the gap.

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dkayiwa: It turns out the framework does not localize these keys, so part of my reasoning above was wrong. In the 2.36.0 sources, RestUtil.wrapErrorResponse just returns the raw exception message, and a wrong code reaches the client as {"error": {"message": "[authentication.error.invalidCodeEntered]", ...}}. The translatedMessage resolution I described only happens for ValidationException via wrapValidationErrorResponse. Removing the handler was still right for the standard envelope, but it means localization has to happen in the O3 frontend by mapping these keys. @jayasanka-sack is that the plan on the frontend side? If the expectation was server-side localized text instead, the scheme would need to resolve the keys through Context.getMessageSourceService() before throwing.

omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Added `enrollmentWindowSeconds` as a configuration property

dkayiwa:
@claude review

omod/src/main/java/org/openmrs/module/authentication/web/WebAuthenticationScheme.java
```diff
@@ -150,4 +152,23 @@ public void afterAuthenticationSuccess(AuthenticationSession session) {
...
+    
+    /**
+     * @param payload the JSON body containing verification details
+     * @param request the current HTTP request
+     * @return a SimpleObject confirming the verification status
+     * @throws UnsupportedOperationException if verification is not supported by the scheme
+     */
+    public SimpleObject verifyEnrollment(SimpleObject payload, HttpServletRequest request) {
+        throw new UnsupportedOperationException("Enrollment verification not supported for " + getSchemeId());
+    }
```
ibacher: General remark: it feels to me that an "EnrollableWebAuthenticationScheme" is something different from a "WebAuthenticationScheme" since baseline authentication schemes (username + password). It seems like instead of extending this class with default functions that throw exceptions, we could just create a new base class for that. This also has the advantage that instead of calling a function and catching an exception, we could use an `instanceof` check and return the same error.

ibacher: I also don't wholly love building an API around `SimpleObject` here at all. It seems like (following the design of this module), we should have a marker interface for a `SecondFactorEnrollmentPayload` that can be differentiated for Totp and Secret question, etc. and we return that. I also don't love just return an unknown `SimpleObject` each enroller feels like it should have one or more `SecondFactorEnrollmentChallenge`.

ibacher: Basically, `SimpleObject` should _never_ be referenced "below" the REST layer.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +228,93 @@ public String getClientName() {
...
+	@Override
+	public SimpleObject initiateEnrollment(HttpServletRequest request) {
+		User user = Context.getAuthenticatedUser();
+		
+		if (user == null) {
+			throw new APIAuthenticationException("authentication.error.mustBeAuthenticatedToEnroll");
+		}
+		
+		if (!isUserConfigurationRequired(user)){
+			throw new IllegalRequestException("authentication.error.secretAlreadyConfigured");
```
ibacher: For similar reasons, we also should not be throwing `IllegalRequestException`s here. That is a web-level exception and should never be referenced inside an `AuthenticationScheme`. Handle this with a module-internal exception and change it at the REST level if necessary.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +228,93 @@ public String getClientName() {
...
+	
+	protected void saveSecret(User user, String secret) {
+		String encryptedSecret = Security.encrypt(secret);
+		
+		AuthenticationScheme twoFactor = AuthenticationConfig.getAuthenticationScheme();
+		if (twoFactor instanceof TwoFactorAuthenticationScheme) {
+			((TwoFactorAuthenticationScheme) twoFactor).addSecondaryAuthenticationSchemeForUser(user, getSchemeId());
+		}
+		
+		Context.getUserService().setUserProperty(user, getSecretUserPropertyName(), encryptedSecret);
```
ibacher, line 314: It would probably be a good idea to factor this out into it's own method. That way we can limit the scope of what we need to mock and avoid the issue with drift.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +228,93 @@ public String getClientName() {
...
+		}
+		
+		boolean isValidCode = verifyCode(temporarySavedSecret, rawCode.toString());
+		if (!isValidCode) {
+			throw new IllegalRequestException("authentication.error.invalidCodeEntered");
+		}
+		
+		saveSecret(user, temporarySavedSecret);
+		
+		request.getSession().removeAttribute(PENDING_ENROLLMENT_SECRET);
```
ibacher, line 304: We should also remove `PENDING_ENROLLMENT_TIME`.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +228,93 @@ public String getClientName() {
...
+		
+		SimpleObject response = new SimpleObject();
+		response.put("isValidCode", isValidCode);
```
ibacher: We know this value is `true` at this point (since we throw before we get here), so maybe we should make that explicit?

ibacher: Granted, I'd even question the value of returning this in the JSON response if the status code is enough to indicate if things worked or not. JSON bodies aren't _required_.

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
```diff
@@ -0,0 +1,55 @@
+package org.openmrs.module.authentication.web.controller;
```
ibacher, line 15: Please remember to add the license header here.

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * REST Controller to manage Two-Factor Authentication enrollment and verification.
+ */
+@Controller
+@RequestMapping(value = "/rest/v1/auth/{schemeId}")
```
ibacher: Use `RestConstants.VERSION_1`, consistent with the rest of the REST API.

omod/src/main/java/org/openmrs/module/authentication/web/WebAuthenticationScheme.java
ibacher: Rethinking `EnrollableWebAuthenticationScheme` a bit... It could also be a `WebAuthenticationScheme` that implements a `EnrollableAuthenticationScheme` interface. That feels cleaner to me and gets the same advantages.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +228,93 @@ public String getClientName() {
...
+		return response;
+	}
+	
+	protected void saveSecret(User user, String secret) {
```
ibacher, line 308: One thought here: it might be better for extensibility to have an orchestration layer (maybe the controller?) that calls `verifyEnrollment` and if that returns `true`, then calls the equivalent of `saveSecret`. We could then move the "register this as a second-factor authentication scheme" part of this method into that orchestration layer, which means that we'd have that already implemented when we extend this to support `EmailAuthenticationScheme` and `SecretQuestionAuthenticationScheme` (we do need to think about the fact that we're aren't _only_ interested in supporting TOTP authentication and ensure that this is designed in an easily-extensible way).

ibacher (changes requested):

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/controller/TwoFactorEnrollmentController.java
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/WebAuthenticationScheme.java
dilankavishka: I created the `EnrollableAuthenticationScheme` interface. `TotpAuthenticationScheme` implements this interface, while base `WebAuthenticationScheme` does not have those methods anymore. and I used the pure Java `Map<String, Object>` and `HashMap`. Only the controller handles `SimpleObject`.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Fixed this with creating an own internal `EnrollmentException` in the core package

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Updated

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: For this, I changed the scheme's verifyEnrollment return type to void and to keep the frontend happy, moved this response generation to the controller, and made it explicitly `true`

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
dilankavishka: Fixed and nowthe `TotpAuthenticationScheme` is only responsible for saving the TOTP secret.

dilankavishka:
Thank you for reviewing my PR @ibacher!
If I summarize all of your comments, most of them are aimed at keeping our architecture reliable and clean (Seperate the interface, decouple the REST layer, etc...). It made me think about handling future edge cases especially making sure our 2FA registration is generic enough to support other upcoming schemes (Email).

Please confirm my thoughts :)

jayasanka-sack:
Good job, @dilankavishka! I have a few small nits we can clean up. I'll send you a DM so we can arrange a quick pair programming session.

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -214,4 +227,86 @@ public String getClientName() {
...
+		saveSecret(user, temporarySavedSecret);
+		
+		// Remove temporary enrollment data from the session once the secret has been verified
+		// and persisted, preventing the same enrollment attempt from being reused.
+		request.getSession().removeAttribute(PENDING_ENROLLMENT_SECRET);
+		request.getSession().removeAttribute(PENDING_ENROLLMENT_TIME);
+	}
+	
+	protected void saveSecret(User user, String secret) {
+		String encryptedSecret = Security.encrypt(secret);
```
dilankavishka, line 309: @ibacher, I have a question about this.

The `Security.encrypt(String) `method we're using here is deprecated. Is there a recommended alternative for encrypting sensitive values like this?

jayasanka-sack (approved):
Thanks for getting this done, @dilankavishka!

PR #55, opened 2026-07-17
AUT-6: Normalize whitespaces from the totp code

dilankavishka:
## Requirements

- [x] My work includes tests or is validated by existing tests.

## Summary
I raise this PR to normalize the verification code by stripping out all whitespaces before validating it. I used `StringUtils.deleteWhitespace` to strip all whitespace from the totp code instead of a simple `trim()` because it only removes leading and trailing whitespace. but if the user inputs "123 456" (which Microsoft Authenticator displays with a space between the 3rd and 4th digit), `trim()` wouldn't do our work. 

## Screenshots
N/A

## Related Issue

https://openmrs.atlassian.net/browse/AUT-6

## Other
N/A

### DIFF
--- a/omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
+++ b/omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
@@ -6,7 +6,6 @@ import org.junit.jupiter.api.Test;
 import org.openmrs.User;
 import org.openmrs.api.APIAuthenticationException;
 import org.openmrs.api.UserService;
-import org.openmrs.api.context.Authenticated;
 import org.openmrs.api.context.AuthenticationScheme;
 import org.openmrs.api.context.BasicAuthenticated;
 import org.openmrs.api.context.Context;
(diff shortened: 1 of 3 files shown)


dilankavishka:
@jayasanka-sack @mogoodrich Could you review this PR please?

omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
```diff
@@ -320,4 +320,10 @@ public User getAuthenticatedUser() {
 			return user;
 		}
 	}
+	
+	@Test
+	public void shouldRemoveSpacesFromTotpVerificationCode() {
```
jayasanka-sack: I think the current test name is a bit misleading because it describes the implementation (removeSpaces) rather than the behavior we're expecting. From a caller's perspective the important part is that verification succeeds even when the input contains spaces, regardless of how that's achieved internally.

some alternatives:
`shouldAcceptVerificationCodeContainingSpaces`
`shouldVerifyTotpCodeWithEmbeddedSpaces`

dilankavishka: I renamed it as `shouldAcceptVerificationCodeContainingSpaces`

jayasanka-sack (approved):
Thanks @dilankavishka!

PR #56, opened 2026-07-21
AUT-28: Fix username display in TOTP authenticator apps

dilankavishka:
## Requirements

- [ ] My work includes tests or is validated by existing tests.

## Summary
This PR fixes an issue where the user's account username is blank in authenticator app after scanning the totp setup qr code.

## Screenshots
N/A

## Related Issue

https://openmrs.atlassian.net/browse/AUT-28

## Other
N/A

### DIFF
--- a/omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
@@ -247,7 +247,7 @@ public class TotpAuthenticationScheme extends WebAuthenticationScheme implements
 		}
 		
 		String secret = generateSecret();
-		String qrCodeUri = generateQrCodeUriForSecret(secret, user.getUsername());
+		String qrCodeUri = generateQrCodeUriForSecret(secret, user.toString());
 		
 		request.getSession().setAttribute(PENDING_ENROLLMENT_SECRET, secret);
 		request.getSession().setAttribute(PENDING_ENROLLMENT_TIME, System.currentTimeMillis());


dilankavishka:
@jayasanka-sack @janithRS Could you review this please?

dkayiwa:
@claude review

jayasanka-sack (approved):
The real issue here is that not every users has user Ids, in those case we need to use the systemId of the user. 

the toString override of the user class is already handling it, so this fix seems appropriate to me. thanks @dilankavishka!

omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java
```diff
@@ -247,7 +247,7 @@ public Map<String, Object> initiateEnrollment(HttpServletRequest request) throws
 		}
 		
 		String secret = generateSecret();
-		String qrCodeUri = generateQrCodeUriForSecret(secret, user.getUsername());
+		String qrCodeUri = generateQrCodeUriForSecret(secret, user.toString());
```
dkayiwa, line 250: I'd spell the fallback out here rather than lean on `toString()`. `User.toString()` is an undocumented debug representation, not an API, so core is free to change what it returns, and if it ever does, every user's authenticator entry silently picks up the new format without anything in this module failing or warning. The module already does the explicit version of this in `UserLogin.getUsername()`. At core 2.7.0 (the module's `require_version`), `toString()` is exactly `isNotBlank(username) ? username : systemId`, so this is behavior-preserving:

```suggestion
		String username = user.getUsername();
		String qrCodeUri = generateQrCodeUriForSecret(secret, StringUtils.isNotBlank(username) ? username : user.getSystemId());
```

(I would have suggested `defaultIfBlank` to mirror `UserLogin`, but this file imports the old `org.apache.commons.lang.StringUtils`, and the module's compile classpath resolves commons-lang 2.4, which predates that method. `isNotBlank` is what this file already uses.)

Not a blocker: as written it does fix the reported case, since the seed admin account has a blank username and system id `admin`.

dkayiwa, line 250: Can this land with a test? Nothing in the suite pins the label down: `shouldInitiateEnrollmentSuccessfully` only asserts `qrCodeUri` is non-null, and its user has a username set, so both the original bug and this fix are invisible to it, and a future refactor could regress it without any signal. The pieces for a real assertion are already on the test classpath (zxing comes in with the java-totp dependency). I ran this locally in `TotpAuthenticationSchemeTest`: it fails on main with `otpauth://totp/?...` and passes on this branch, so feel free to lift it as-is:

```java
@Test
public void shouldUseSystemIdAsQrCodeLabelWhenUsernameIsBlank() throws Exception {
	candidateUser.setUsername("");
	candidateUser.setSystemId("admin");
	Context.setUserContext(new MockUserContext(candidateUser));

	Map<String, Object> result = authenticationScheme.initiateEnrollment(request);

	String qrCodeUri = (String) result.get("qrCodeUri");
	byte[] png = java.util.Base64.getDecoder().decode(qrCodeUri.substring(qrCodeUri.indexOf(',') + 1));
	java.awt.image.BufferedImage image = javax.imageio.ImageIO.read(new java.io.ByteArrayInputStream(png));
	String otpauthUri = new com.google.zxing.MultiFormatReader().decode(new com.google.zxing.BinaryBitmap(
			new com.google.zxing.common.HybridBinarizer(
					new com.google.zxing.client.j2se.BufferedImageLuminanceSource(image)))).getText();
	assertThat(otpauthUri, org.hamcrest.Matchers.startsWith("otpauth://totp/admin?"));
}
```

PR #57, opened 2026-08-05
AUT-30: Fix secondary authentication not saving to DB

dilankavishka:
This PR is going to fix the bug where secondary authentication scheme (TOTP) fail to save to the database.

Ticket: https://openmrs.atlassian.net/browse/AUT-30

### DIFF
--- a/omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
@@ -330,7 +330,7 @@ public class TwoFactorAuthenticationScheme extends WebAuthenticationScheme {
 		List<String> schemeIds = getSecondaryAuthenticationSchemeIdsForUser(user);
 		if (!schemeIds.contains(schemeId)) {
 			schemeIds.add(schemeId);
-			Context.getUserService().setUserProperty(user, USER_PROPERTY_SECONDARY_TYPE, String.join(",", schemeId));
+			Context.getUserService().setUserProperty(user, USER_PROPERTY_SECONDARY_TYPE, String.join(",", schemeIds));
 		}
 	}
 
--- a/omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
+++ b/omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
@@ -330,7 +330,7 @@ public class TwoFactorAuthenticationScheme extends WebAuthenticationScheme {
 		List<String> schemeIds = getSecondaryAuthenticationSchemeIdsForUser(user);
 		if (!schemeIds.contains(schemeId)) {
 			schemeIds.add(schemeId);
-			user.setUserProperty(USER_PROPERTY_SECONDARY_TYPE, String.join(",", schemeIds));
+			Context.getUserService().setUserProperty(user, USER_PROPERTY_SECONDARY_TYPE, String.join(",", schemeId));
 		}
 	}
 


dilankavishka:
@jayasanka-sack Could you review this please?

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
```diff
@@ -330,7 +330,7 @@ public void addSecondaryAuthenticationSchemeForUser(User user, String schemeId)
 		List<String> schemeIds = getSecondaryAuthenticationSchemeIdsForUser(user);
```
dkayiwa, line 330: A user who enrolled before this fix already has `authentication.totp.secret` persisted but no `authentication.secondaryType`, and this change doesn't move them out of that state. `getSecondaryAuthenticationSchemeIdsForUser` still returns empty for them, so they're never challenged for TOTP, and they can't re-enroll either: `isUserConfigurationRequired` sees the stored secret, so both `initiateEnrollment` and `verifyEnrollment` throw `secretAlreadyConfigured`. There's no unenroll endpoint, so the only way out is clearing the user property by hand.

The #53 endpoints aren't in any released tag yet (`git tag --contains 419e3cf` comes back empty), so this is a dev/QA problem rather than a data migration. It will still bite whoever verifies this fix on an environment where they already enrolled. Is a reset or unenroll path planned as a follow-up?

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
```diff
@@ -330,7 +330,7 @@ public void addSecondaryAuthenticationSchemeForUser(User user, String schemeId)
...
 		if (!schemeIds.contains(schemeId)) {
 			schemeIds.add(schemeId);
-			user.setUserProperty(USER_PROPERTY_SECONDARY_TYPE, String.join(",", schemeIds));
+			Context.getUserService().setUserProperty(user, USER_PROPERTY_SECONDARY_TYPE, String.join(",", schemeIds));
```
dkayiwa, line 333: This should come with a test, and the one that already covers this line can't be it. `TotpAuthenticationSchemeTest.shouldVerifyEnrollmentSuccessfully` drives the controller and then asserts `candidateUser.getUserProperty("authentication.secondaryType")` is `"totp"`, but the `UserService` proxy its `setup()` installs implements `setUserProperty` as `u.setUserProperty(k, v)`. That collapses the in-memory write and the service write into the same observable result, so the assertion passed with the bug in place. I ran the class on both branches: 17/17 green either way.

What does separate them is asserting the write went through `UserService`. Recording the keys in the existing proxy and checking for the secondary-type key fails on `main` (`Expected: a collection containing "authentication.secondaryType" but: was empty`) and passes with your change:

```java
// field on the test class
final List<String> persistedPropertyKeys = new ArrayList<>();

// inside the setup() proxy, next to u.setUserProperty(k, v)
persistedPropertyKeys.add(k);

// in shouldVerifyEnrollmentSuccessfully
assertThat(persistedPropertyKeys, hasItem(TwoFactorAuthenticationScheme.USER_PROPERTY_SECONDARY_TYPE));
```

I'd add that (or an equivalent) here rather than as a follow-up, since otherwise the same one-line regression can ship green again.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
```diff
@@ -330,7 +330,7 @@ public void addSecondaryAuthenticationSchemeForUser(User user, String schemeId)
...
 		}
 	}
```
dkayiwa, line 336: `setSecondaryAuthenticationSchemeIdsForUser` just below still writes through `user.setUserProperty`, so it carries the same defect this PR is fixing. Nothing calls it today, it's been unused since it was added in #45, so nothing is broken right now, but the next caller re-creates AUT-30. The same one-line change would keep the two setters consistent.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dilankavishka: Yes! the unenroll/remove path comes as a followup work.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dilankavishka: Removed that completely.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dilankavishka: Updated the test

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dilankavishka: For now, to test this PR on an already enrolled account, you can manually clear your state by deleting the `authentication.totp.secret` from your User Properties. That will let you trigger the enrollment flow again.

dkayiwa:
@claude review

dilankavishka:
@dkayiwa @jayasanka-sack @janithrs Could you review this PR please?

I have reverted the changes in the unit test and added `TwoFactorAuthenticationSchemeIntegrationTest` specifically for this scenario.

Note: I only added this integration test for this specific scenario to get this bug fix merged. I recognize that we need to improve the test suite with more integration tests, but my plan is to write those in a separate, dedicated PR so this one doesn't grow too large.

dkayiwa:
Can you look into the above build failures?

dilankavishka:
> Can you look into the above build failures?

I'm working on it @dkayiwa :D

wikumChamith (commented):
I think the unit tests may be leaving an `AuthenticationConfig` in Context's static `authenticationScheme` field, and in the thread-local `UserContext`. This causes the Swing credentials dialog to open, and on a machine without a display like our CI, we get a `HeadlessException`. Adding this to the failing test should work:

```java
@BeforeAll
static void resetAuthenticationScheme() throws Exception {
    AuthenticationConfig.setConfig(new Properties());
    Field field = Context.class.getDeclaredField("authenticationScheme");
    field.setAccessible(true);
    field.set(null, new UsernamePasswordAuthenticationScheme());
    Context.clearUserContext();
}
```

dilankavishka:
> ```java
> AuthenticationConfig.setConfig(new Properties());
>     Field field = Context.class.getDeclaredField("authenticationScheme");
>     field.
> ```

Thank you so much @wikumChamith!!!

dilankavishka:
@dkayiwa Could you please take a look in to this again?

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+	@Nested
+	@DisplayName("addSecondaryAuthenticationSchemeForUser")
+	class AddSecondaryAuthenticationSchemeForUser {
+		
+		@Test
+		@DisplayName("should save the scheme to the database")
+		void shouldSaveSchemeToDatabase() {
+			User user = Context.getUserService().getUser(1);
+			
+			scheme.addSecondaryAuthenticationSchemeForUser(user, "totp");
```
dkayiwa, line 69: As written this test cannot fail, so merging it leaves AUT-30 unguarded: whoever turns `addSecondaryAuthenticationSchemeForUser` back into an in-memory write gets a green build. I reverted that one line, rebuilt clean (the bytecode does call `org/openmrs/User.setUserProperty`), and ran the omod suite: 128/128 green, this test included.

`getUser(1)` hands back a Hibernate-managed instance, so mutating its properties map marks it dirty and `Context.flushSession()` writes it out whether or not `UserService` was involved. In production the user comes from `Context.getAuthenticatedUser()`, which is detached across requests, and that is why the write went missing.

Detaching first reproduces it. I ran this version both ways: it fails on the old code with `expected: <totp> but was: <>`, and passes with your change.

```suggestion
			User user = Context.getUserService().getUser(1);
			user.getUserProperties().size(); // initialise the lazy collection before detaching
			Context.evictFromSession(user);
			
			scheme.addSecondaryAuthenticationSchemeForUser(user, "totp");
```

The `getUserProperties()` call is there because the map is lazy; without it the test errors with `LazyInitializationException` instead of asserting anything.

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+public class TwoFactorAuthenticationSchemeIntegrationTest extends BaseModuleWebContextSensitiveTest {
+	
+	@BeforeAll
+	static void resetAuthenticationScheme() throws Exception {
+		AuthenticationConfig.setConfig(new Properties());
+		Field field = Context.class.getDeclaredField("authenticationScheme");
+		field.setAccessible(true);
+		field.set(null, new UsernamePasswordAuthenticationScheme());
+		Context.clearUserContext();
+	}
```
dkayiwa, line 49: This would do more good in `BaseAuthenticationTest.teardown()`. What it undoes is the state that `setAuthenticationSchemeOnContext()` in that base class sets and never puts back, so every context-sensitive test added to this module from here on will hit the same `HeadlessException` and need this block copied into it. It is a confusing failure to land on, too, since it reads as a CI environment problem rather than test pollution.

I reproduced it locally with `-Dsurefire.runOrder=alphabetical -Djava.awt.headless=true`, which sorts this class last, right after `TwoFactorRememberMeTest`. Moving the reset into `teardown()` and deleting this `@BeforeAll` left the suite green in that same order.

Not blocking, and reasonable as a follow-up if you would rather keep this PR tight.

omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
```diff
@@ -44,7 +44,7 @@ public class TotpAuthenticationSchemeTest extends BaseWebAuthenticationTest {
 	MockTotpAuthenticationScheme authenticationScheme;
 	User candidateUser;
 	UserLogin userLogin;
-
+
```
dkayiwa: This line picked up a trailing tab and nothing else, which pulls the whole file into the diff for no reason. Worth reverting so the change stays to the three files it needs.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dkayiwa: Still gone at 77d4768, and I checked it independently rather than take the bot's word for it. authenticationui's `TwoFactorSetupPageController` calls it at line 117 on main, `twoFactorSetup.gsp` is a live page reachable from the user account page, and that module's config.xml requires `org.openmrs.module.authentication` with no version bound. So a distro that picks up 2.4.0 under today's authenticationui build starts fine and then throws `NoSuchMethodError` the first time someone saves their two-factor settings. Please put the method back, with the same `Context.getUserService().setUserProperty(...)` body you used in the setter above.

omod/src/test/java/org/openmrs/module/authentication/web/TotpAuthenticationSchemeTest.java
dilankavishka: Done

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
dilankavishka: Updated

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
dilankavishka: Thank you for pointing this out. My plan is to improve the integration test suite in this module within my next PR. shall we leave this to that PR? (With this ticket: https://openmrs.atlassian.net/browse/AUT-31)

dilankavishka: Thank you for pointing this out. My plan is to improve the integration test suite in this module within my next PR. shall we leave this to that PR? (With this ticket: https://openmrs.atlassian.net/browse/AUT-31)

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dkayiwa: Still gone at 67405d5, and this is the one thing left before merge. I checked it against the released omod this time rather than authenticationui's main branch, so the "they'll fix it when they bump the version" reading doesn't apply: `authenticationui-omod-1.3.0.jar` carries `invokevirtual org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.setSecondaryAuthenticationSchemeIdsForUser:(Lorg/openmrs/User;Ljava/util/List;)V` in `TwoFactorSetupPageController`, and its config.xml requires `org.openmrs.module.authentication` with no version bound. So 1.3.0 loads fine alongside 2.4.0 and then throws `NoSuchMethodError`.

Being more precise than I was above about which clicks break, since a fresh TOTP enrollment redirects to the configuration page before reaching that line: it is the "Remove" and "Set preferred" links on twoFactorSetup.page. Both POST with `schemeId` set and fall through to line 117, and neither has an equivalent in `addSecondaryAuthenticationSchemeForUser`, which can only append.

Restoring it with the body you used in the setter above keeps the API intact, and it also finishes the sweep, since after your change this is the only user property write left in the module that doesn't go through `UserService`:

```java
	public void setSecondaryAuthenticationSchemeIdsForUser(User user, List<String> schemeIds) {
		Context.getUserService().setUserProperty(user, USER_PROPERTY_SECONDARY_TYPE, String.join(",", schemeIds));
	}
```

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
dkayiwa: Yes, AUT-31 is a fine home for the refactor.

While the block stays here, could you add a line saying what it is for? I checked whether it was still earning its place now that `BaseAuthenticationTest` sets the config in memory, and it is: delete it and `mvn -pl omod test -Dsurefire.runOrder=alphabetical` dies with `java.awt.HeadlessException`, while with it the omod suite is 128/128 in both that order and the default one. Without a note it reads like leftover scaffolding, and the exception you get after deleting it points at the display rather than at test pollution.

dilankavishka: Done :)

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dkayiwa: `javap` on the released `authentication-omod-2.3.0.jar` lists `setSecondaryAuthenticationSchemeIdsForUser(User, List)` among the public methods, so it is a shipped part of this module's API rather than something authenticationui only compiled against a branch. And the current `authenticationui-omod-1.4.0-SNAPSHOT` (deployed 2026-08-03) still carries the `invokevirtual` for it, so there is no migration in flight upstream that would land ahead of 2.4.0.

Still gone at a36e952, and still the only thing I have outstanding on this PR. Merged as is, a distro that picks up 2.4.0 under any authenticationui built to date throws `NoSuchMethodError` when someone saves their two-factor settings, because that module's config.xml requires `org.openmrs.module.authentication` with no version bound and so loads against 2.4.0 without complaint.

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
dkayiwa: `user.getUserProperties().size()` on a line of its own reads as a no-op, so it is a fair bet that whoever tidies this test next deletes it. I removed it and reran: `org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: org.openmrs.User.userProperties, could not initialize proxy - no Session`, thrown out of the `addSecondaryAuthenticationSchemeForUser` call, since that reads the property map off the detached user.

A trailing `// initialise the lazy collection before detaching` is enough to stop that. Not blocking.

dilankavishka: Updated.

omod/src/main/java/org/openmrs/module/authentication/web/TwoFactorAuthenticationScheme.java
dilankavishka: Updated

dilankavishka:
@dkayiwa Could you please take a look in to this again?

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+			
+			Context.flushSession();
+			Context.clearSession();
+			
+			User savedUser = Context.getUserService().getUser(1);
+			
+			String savedProperty = savedUser.getUserProperty(TwoFactorAuthenticationScheme.USER_PROPERTY_SECONDARY_TYPE);
+			assertEquals("totp", savedProperty);
+		}
+	}
```
dkayiwa, line 100: `setSecondaryAuthenticationSchemeIdsForUser` could use a test of its own, though I would not hold the merge for it. I reverted that one line back to `user.setUserProperty(...)`, left everything else alone, rebuilt, and the suite is still green: 128/128 in omod, 15/15 in api. So half of this fix can be undone without CI noticing, which is what the nested class above rules out for the other setter.

A second nested class next to it closes that. I ran it both ways: it fails with `expected: <totp> but was: <>` on the in-memory write and passes with your version, and omod goes to 129/129.

```java
	@Nested
	@DisplayName("setSecondaryAuthenticationSchemeIdsForUser")
	class SetSecondaryAuthenticationSchemeIdsForUser {

		@Test
		@DisplayName("should save the schemes to the database")
		void shouldSaveSchemesToDatabase() {
			User user = Context.getUserService().getUser(1);
			user.getUserProperties().size(); // initialise the lazy collection before detaching
			Context.evictFromSession(user);

			scheme.setSecondaryAuthenticationSchemeIdsForUser(user, Collections.singletonList("totp"));

			Context.flushSession();
			Context.clearSession();

			User savedUser = Context.getUserService().getUser(1);

			String savedProperty = savedUser.getUserProperty(TwoFactorAuthenticationScheme.USER_PROPERTY_SECONDARY_TYPE);
			assertEquals("totp", savedProperty);
		}
	}
```

It needs `import java.util.Collections;` at the top.

dilankavishka: Done

api/src/test/java/org/openmrs/module/authentication/BaseAuthenticationTest.java
```diff
@@ -68,7 +68,7 @@ protected void setRuntimeProperties(Properties p) {
 		p.setProperty(OpenmrsConstants.APPLICATION_DATA_DIRECTORY_RUNTIME_PROPERTY, appDataDir.getAbsolutePath());
 		OpenmrsUtil.storeProperties(p, runtimePropertiesFile, "test");
 		Context.setRuntimeProperties(p);
-		AuthenticationConfig.reloadConfigFromRuntimeProperties("openmrs");
+		AuthenticationConfig.setConfig(AuthenticationUtil.getPropertiesWithPrefix(p, AuthenticationConfig.PREFIX, false));
```
dkayiwa, line 71: This line was the last thing in the build that ran `reloadConfigFromRuntimeProperties`, and nothing exercises it now. I checked by stubbing that method's body out to `config = new Properties()`: on `main` it fails 95 of the 127 omod tests, on this branch the build stays green (15 in api, 129 in omod). All of that coverage was incidental, coming from the omod scheme tests going through `setRuntimeProperties`, and the api module was green under the stub on `main` too.

The method is still live production code, `AuthenticationFilter` line 121 calls it on every unauthenticated request when `authentication.settings.cached` is false. Its javadoc calls that a development and testing setting rather than a production one, so this isn't a hole in a deployed system, but a regression in it would now ship green and land on whoever is debugging authentication config, in the shape of their `authentication.*` settings silently not applying.

To be clear, I wouldn't go back to the disk round trip. Reading `p` directly is the better fixture, and the api module is clear of the pollution that made this change necessary. A direct test in `AuthenticationConfigTest` covers what gets dropped. I ran it both ways: it fails with `Expected: "basic" but: was null` against the stub, passes as-is, and takes api to 16. It needs no new imports.

```java
	@Test
	public void shouldReloadConfigFromRuntimeProperties() {
		Properties p = new Properties();
		p.setProperty(SCHEME, "basic");
		p.setProperty("some.unrelated.property", "ignored");
		setRuntimeProperties(p);
		AuthenticationConfig.setConfig(new Properties());
		AuthenticationConfig.reloadConfigFromRuntimeProperties("openmrs");
		assertThat(AuthenticationConfig.getProperty(SCHEME), equalTo("basic"));
		assertThat(AuthenticationConfig.getKeys().contains("some.unrelated.property"), equalTo(false));
	}
```

Not blocking.

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
dkayiwa: The comment you added in b2890ec is gone again at the head commit, so this block is back to reading like leftover scaffolding. It looks like collateral from `Add a new test case` rather than a deliberate removal: that same commit also left an extra blank line above the `addSecondaryAuthenticationSchemeForUser` call, which the second nested class doesn't have. Neither blocks merge, but both are worth tidying while you're in the file.

dilankavishka: My bad. Sorry for the mistake @dkayiwa and thank you for the patience.

api/src/test/java/org/openmrs/module/authentication/BaseAuthenticationTest.java
dilankavishka: Updated

omod/src/test/java/org/openmrs/module/authentication/web/integration/TwoFactorAuthenticationSchemeIntegrationTest.java
dkayiwa: The comment block sits at column 0 rather than one tab in like the rest of the class body, so the javadoc reads as though it documents the file instead of `resetAuthenticationScheme`. Indenting it one tab, as it was at b2890ec, is all it needs.

The stray blank line above the `addSecondaryAuthenticationSchemeForUser` call is still there too: line 68 is three tabs of whitespace and line 69 is empty, where the second nested class has a single separator. Both are cosmetic, so neither holds up the merge.

dilankavishka: Fixed

dilankavishka:
@dkayiwa could we get this PR merged if everything looks good to you?

